PR #2, opened 2017-05-05
Adding 2 new Layout tests and few minor fixes

ssivov:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/clarity-js/2)

### DIFF
--- a/src/layout/layout.ts
+++ b/src/layout/layout.ts
@@ -17,7 +17,7 @@ class Layout implements IComponent {
   private mutationSequence: number;
   private domDiscoverComplete: boolean;
   private domDiscoverQueue: number[];
-  private domPreDiscoverMutations: IMutationBatch[];
+  private domPreDiscoverMutations: MutationRecord[][];
   private originalProperties: INodePreUpdateInfo[];
 
   public reset(): void {
@@ -149,10 +149,7 @@ class Layout implements IComponent {
     this.domDiscoverComplete = true;
 
     if (this.domPreDiscoverMutations.length > 0) {
-      let allMutationRecords: MutationRecord[] = [];
-      for (let i = 0; i < this.domPreDiscoverMutations.length; i++) {
-        allMutationRecords = allMutationRecords.concat(this.domPreDiscoverMutations[i].mutations);
-      }
+      let allMutationRecords = Array.prototype.concat.apply([], this.domPreDiscoverMutations);
       this.mutation(allMutationRecords, Math.round(getPageContextBasedTimestamp()));
     }
 
@@ -174,10 +171,7 @@ class Layout implements IComponent {
           break;
       }
     }
-    this.domPreDiscoverMutations.push({
-      time,
-      mutations
-    });
+    this.domPreDiscoverMutations.push(mutations);
   }
 
   // Looks at whether some node's attributes/characterData have mutated for the first time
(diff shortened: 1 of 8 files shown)


src/layout/layout.ts
```diff
@@ -68,7 +68,7 @@ class Layout implements IComponent {
   // them with real data asynchronously (if it takes too long to do at once) by yielding a thread
   // and returning to it later through a set timeout
   private discoverDom() {
-    let discoverTime = getPageContextBasedTimestamp();
+    let discoverTime = Math.round(getPageContextBasedTimestamp());
```
toby-walker: Would it be cleaner to provide a rounded timestamp routine rather than rounding in the client?

src/layout/layout.ts
```diff
@@ -148,11 +148,13 @@ class Layout implements IComponent {
   private onDomDiscoverComplete() {
     this.domDiscoverComplete = true;
 
-    let allMutationRecords: MutationRecord[] = [];
-    for (let i = 0; i < this.domPreDiscoverMutations.length; i++) {
-      allMutationRecords = allMutationRecords.concat(this.domPreDiscoverMutations[i].mutations);
+    if (this.domPreDiscoverMutations.length > 0) {
+      let allMutationRecords: MutationRecord[] = [];
+      for (let i = 0; i < this.domPreDiscoverMutations.length; i++) {
```
toby-walker: would concat.apply(this, domPreDiscovreMutations) be faster and simpler? concat() can concatenate multiple arrays.

ssivov: Yes, it looks like a good shortcut - will make a change

src/layout/layout.ts
ssivov: My personal preference is to keep the function clean and doing exactly what it says. Also, doing so won't eliminate all Math.round() calls in the code base, because some timestamps come to us from, for example, performance APIs, so we still have to manually round them.

sarveshnagpal:
Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/layout/layout.ts, line 71 at r1](https://reviewable.io:443/reviews/microsoft/clarity-js/2#-KjaMyBD8lT3TGbodHgj-r1-71:-KjaNw6Czx-cw-PRsj9N:br9q0lx) ([raw file](https://github.com/microsoft/clarity-js/blob/64004eb3a2890005ac61a1e4cf738f57db2010d3/src/layout/layout.ts#L71)):*
<details><summary><i>Previously, ssivov wrote…</i></summary><blockquote>

My personal preference is to keep the function clean and doing exactly what it says. Also, doing so won't eliminate all Math.round() calls in the code base, because some timestamps come to us from, for example, performance APIs, so we still have to manually round them.
</blockquote></details>

Ideally, shouldn't getPageContextBasedTimestamp always be millisecond precision? Anytime we need to compute time deltas, we can fallback to unix time function so we don't lose precision. On your second point, I agree we will still need to round off certain timestamps but hopefully never calls to this specific function.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/clarity-js/2)*

sarveshnagpal:
Reviewed 3 of 4 files at r1, 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/clarity-js/2)*

src/layout/layout.ts
toby-walker: I'm not suggesting you add the round() to this function but create a separate function that also does the rounding and returns that rounded time rather than putting that multiple places in your code.

PR #4, opened 2017-05-16
ShadowDom sanity checks

ssivov:
Ensuring that rare inconsistency occurrences in the shadowDom don't result in JavaScript errors

### DIFF
--- a/src/layout/layout.ts
+++ b/src/layout/layout.ts
@@ -342,19 +342,23 @@ class Layout implements IComponent {
   }
 
   private mutation(mutations: MutationRecord[], time: number) {
-    let shadowDomConsistentBeforeMutations = this.shadowDomConsistent;
-    let summary = this.shadowDom.applyMutationBatch(mutations);
+    // Don't process mutations on top of the inconsistent state.
+    // ShadowDom mutation processing logic requires consistent state as a prerequisite.
+    // If we end up in the inconsistent state, that means that something went wrong already,
+    // so we can give up on the following mutations and should investigate the cause of the error.
+    // Continuing to process mutations can result in javascript errors and lead to even more inconsistencies.
+    if (this.shadowDomConsistent) {
+      let summary = this.shadowDom.applyMutationBatch(mutations);
 
-    // Make sure ShadowDom arrived to the consistent state
-    if (shadowDomConsistentBeforeMutations) {
+      // Make sure ShadowDom arrived to the consistent state
       this.shadowDomConsistent = this.shadowDom.mirrorsRealDom();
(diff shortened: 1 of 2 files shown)


src/layout/layout.ts
```diff
@@ -342,19 +342,23 @@ class Layout implements IComponent {
...
+      assert(this.shadowDomConsistent, "mutation");
       if (!this.shadowDomConsistent) {
         debug(`>>> ShadowDom doesn't match PageDOM after mutation batch #${this.mutationSequence}!`);
       }
-    }
 
-    this.processMutations(summary, time);
+      this.processMutations(summary, time);
+    }
     this.mutationSequence++;
```
toby-walker: Two very minor points: 

1. Just return when !this.shadowDomConsistent
2. Should the this.mutationSequence++ but within the scope of the check if you don't do 1?

src/layout/layout.ts
```diff
@@ -342,19 +342,23 @@ class Layout implements IComponent {
...
-    let summary = this.shadowDom.applyMutationBatch(mutations);
-
-    // Make sure ShadowDom arrived to the consistent state
-    if (shadowDomConsistentBeforeMutations) {
+    // Don't process mutations on top of the inconsistent state.
+    // ShadowDom mutation processing logic requires consistent state as a prerequisite.
+    // If we end up in the inconsistent state, that means that something went wrong already,
+    // so we can give up on the following mutations and should investigate the cause of the error.
+    // Continuing to process mutations can result in javascript errors and lead to even more inconsistencies.
+    if (this.shadowDomConsistent) {
```
toby-walker: Do we send an event when this happens (i.e. we become inconsistent)?

src/layout/shadowdom.ts
```diff
@@ -159,21 +168,24 @@ export class ShadowDom {
   }
 
   public mirrorsRealDom(): boolean {
```
toby-walker: Where is the test case for the inconsistent DOM? Wasn't clear from comment or test name.

src/layout/layout.ts
ssivov: We do, there is a failed assertion for it.

src/layout/shadowdom.ts
ssivov: There isn't one, because i'm not aware of any exact scenarios right now that would lead to inconsistencies. However, looking at our data might help us figure out the root causes of the errors and whether they result from shadowDom arriving to the inconsistent state or something else.

src/layout/layout.ts
toby-walker: Resolved

PR #6, opened 2017-05-18
Setting up Clarity as the NPM package

ssivov:


src/wireup.ts
```diff
@@ -1,8 +1,14 @@
 import { config } from "./config";
 import { activate, bind, teardown } from "./core";
 
-if (config.activateEvent) {
-  bind(window, config.activateEvent, activate);
-} else {
-  setTimeout(activate, 0);
+export function start() {
+  if (config.activateEvent) {
```
sarveshnagpal: Let's keep it simple and simply expose activate here. 
Also delete activateEvent - that's for consumers of this library to decide. 

Finally, we should allow a config object here to override the upload URL or callback handler.

PR #7, opened 2017-05-19
Adding support for custom send handler through config

ssivov:


clarity.d.ts
```diff
@@ -25,6 +25,10 @@ interface IConfig {
   // Send back instrumentation data, if set to true
   instrument?: boolean;
 
+  // Pointer to the function which would be responsible for sending the data
+  // If left unspecified, raw payloads will be uploaded to the uploadUrl endpoint
+  sendCallback?: (payload: string, onSuccess?: (status: number) => void, onFailure?: (status: number) => void) => void;
```
toby-walker: Should you add some sort of unique identifier for tracking the request? Otherwise how do you know which failed (unless you use a unique callback per invocation).

src/core.ts
```diff
@@ -169,8 +169,9 @@ function uploadDroppedPayloadsMappingFunction(sequenceNumber: string, droppedPay
 }
 
 function upload(payload: string, onSuccess?: (status: number) => void, onFailure?: (status: number) => void) {
-  // Send to the backend
-  if (config.uploadUrl.length > 0) {
+  if (config.sendCallback) {
+    config.sendCallback(payload, onSuccess, onFailure);
+  } else if (config.uploadUrl.length > 0) {
```
toby-walker: Package up the default case as a function so that it's cleaner. You can even reference it in the default configuration.

test/core.ts
```diff
@@ -18,12 +18,26 @@ describe("Functional Tests", () => {
...
   afterEach(cleanupFixture);
 
-  it("should validate that modules work fine together", (done) => {
+  it("validates that modules work fine together", (done) => {
     let events = getAllSentEvents();
     assert.equal(events.length >= 10, true);
     done();
   });
 
+  it("validates that custom sendCallback is invoked when passed through config", (done) => {
```
toby-walker: Add tests for the on success and on failure cases.

clarity.d.ts
ssivov: onSuccess and onFailure functions wrap the payload, so we know which data (event id's, etc) was not delivered.

src/core.ts
ssivov: Done

toby-walker: There's only one commit in the PR. Maybe you didn't push this change?

test/core.ts
ssivov: Added pretty thorough tests

src/core.ts
ssivov: Looks like my terminal hung on push. Updated the branch now.

src/core.ts
```diff
@@ -169,18 +177,22 @@ function uploadDroppedPayloadsMappingFunction(sequenceNumber: string, droppedPay
 }
 
 function upload(payload: string, onSuccess?: (status: number) => void, onFailure?: (status: number) => void) {
-  if (config.sendCallback) {
-    config.sendCallback(payload, onSuccess, onFailure);
+  if (config.uploadHandler) {
+    config.uploadHandler(payload, onSuccess, onFailure);
   } else if (config.uploadUrl.length > 0) {
```
ssivov: Makes sense, i moved it to the handler.

clarity.d.ts
```diff
@@ -27,7 +27,7 @@ interface IConfig {
 
   // Pointer to the function which would be responsible for sending the data
   // If left unspecified, raw payloads will be uploaded to the uploadUrl endpoint
-  sendCallback?: (payload: string, onSuccess?: (status: number) => void, onFailure?: (status: number) => void) => void;
+  uploadHandler?: (payload: string, onSuccess?: (status: number) => void, onFailure?: (status: number) => void) => void;
```
ssivov: Great tip, thanks!

src/core.ts
```diff
@@ -92,10 +92,18 @@ export function addEvent(type: string, eventState: any, time?: number) {
   }
   nextPayload.push(eventStr);
   nextPayloadLength += eventStr.length;
-  if (timeout) {
-    clearTimeout(timeout);
+
+  // Edge case:
+  // Don't reschedule upload when next payload consists of exactly one XhrError instrumentation event.
+  // This helps us avoid the infinite loop in the case when all requests fail (e.g. dropped internet connection)
```
ssivov: This is not for the retries, but rather for sending new payload. When some xhr fails, we store its payload for retry, but we generate a new instrumentation event, which becomes a new payload of its own. Then, this new payload consisting of just this event is scheduled to be sent through a timeout, and, if it fails to deliver as well, we get into a loop.

PR #9, opened 2017-05-26
Samart/add error component

ender336:
Bringing in a basic error tracking to plug into clarity, can be called directly or can rely on it listening to unhandled exceptions on the page.


src/apicheck.ts
```diff
@@ -13,6 +13,7 @@ let apiContainer = {
   optional: [
     "performance",
     "performance.getEntriesByType",
+    "window.addEventListener"
   ]
```
toby-walker: What about older IE browser (pre IE9) that don't have this? Should you use the equivalent call (EventTarget.attachEvent ) using feature detection or should we just not target those old browsers?

src/errors.ts
```diff
@@ -0,0 +1,82 @@
+import { register } from "./core";
+import { instrument } from "./instrumentation";
+
+let regexEscape = (str) => str.replace(/([.?*+^$&[\]\\(){}|<>-])/g, "\\$1");
+
+export function logError(errorToLog) {
+    // if errorToLog["error"] doesn't exist, occasionally we can get information directly from e
```
toby-walker: Minor: What is "e" referring to here?

src/errors.ts
```diff
@@ -0,0 +1,82 @@
...
+    let error = errorToLog["error"] || errorToLog;
+    // this is a default value to indicate the object passed to us had no message member
+    let defaultMessage = "noMessage";
+    let source = errorToLog["filename"];
+    let lineno = errorToLog["lineno"];
+    let colno = errorToLog["colno"];
+    let message = error.message || defaultMessage;
```
toby-walker: Why have this default? Just leave it null if there is nothing.

src/errors.ts
```diff
@@ -0,0 +1,82 @@
...
+    let stack = error.stack;
+    let hrefRegex = new RegExp(regexEscape(location.href), "g");
```
toby-walker: Don't use regex. Use encodeURI() or encodeURIComponent() instead.

src/errors.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    if (stack) {
+        // ---------------------------------------------------------------------------------------------------------------------
+        // the stack has the potential to be very long with a lot of repeated filenames. The following is a small compression to
```
toby-walker: Maybe but this looks risky and ad hoc to do on the client. Wouldn't it be better to send over as is and post-process it on the data side?

src/errors.ts
```diff
@@ -0,0 +1,82 @@
...
+    public activate() {
+        window.addEventListener("error", logError, false);
+    }
+
+    public reset(): void {
+        return;
+    }
+
+    public teardown() {
+        return;
```
toby-walker: Shouldn't you remove the error listener?

test/apicheck.ts
```diff
@@ -17,7 +17,7 @@ describe("ApiCheck Tests", () => {
   it("validates that missing api event is sent when required api is missing", (done) => {
     core.teardown();
 
-    // Function.prorotype.bind is a required API for Clarity to work
+    // Function.protoype.bind is a required API for Clarity to work
```
ssivov: Almost! Should be 'prototype' though :)

src/errors.ts
```diff
@@ -0,0 +1,82 @@
+import { register } from "./core";
+import { instrument } from "./instrumentation";
+
+let regexEscape = (str) => str.replace(/([.?*+^$&[\]\\(){}|<>-])/g, "\\$1");
+
+export function logError(errorToLog) {
```
ssivov: Why are you exporting this function? Seems like it should be ErrorMonitor's internal routine

test/errors.ts
```diff
@@ -0,0 +1,97 @@
...
+        let message = "sample error text";
+        let filename = "sample filename";
+        let lineno = "sample error text";
+        let colno = "sample error text";
+        syntheticEvent.initEvent("error", true, true);
+        syntheticEvent["message"] = message;
+        syntheticEvent["filename"] = filename;
+        syntheticEvent["lineno"] = lineno;
+        syntheticEvent["colno"] = colno;
+        errors.logError(syntheticEvent);
```
ssivov: Should you have the window dispatch this event, rather than invoking a function directly ?

test/errors.ts
```diff
@@ -0,0 +1,97 @@
...
+        let errorEvents = events.filter((event) => event.state.type === Instrumentation.JsError);
+        assert.equal(errorEvents.length, 1);
+        assert.equal(errorEvents[0].type, InstrumentationEventName);
+        assert.equal(errorEvents[0].state.type, Instrumentation.JsError);
+        assert.equal(errorEvents[0].state.message, message);
+        assert.equal(errorEvents[0].state.source, filename);
+        assert.equal(errorEvents[0].state.lineno, lineno);
+        assert.equal(errorEvents[0].state.colno, colno);
+        done();
+        }
```
ssivov: super nit:
can you please combine 36&37 into '});'

src/errors.ts
ender336: I'm not very worried about the risk - the same compression has been running on Bing for a year or so. There's an argument to be made that the space saving is trivial compared to the amount of data the Clarity is already sending though - let me check what sorts of savings I'm actually getting and see what % of the payload they are.

ender336: yeah it doesn't look like the saving is worthwhile - I'll come back to a compression later if we decide we want to shrink the payloads significantly

src/apicheck.ts
ender336: Will do

test/apicheck.ts
ender336: haha wow

test/errors.ts
ender336: phantomjs can't handle dispatching events that aren't keyboard or mouse :(

test/errors.ts
ender336: sure

src/apicheck.ts
```diff
@@ -13,6 +13,7 @@ let apiContainer = {
   optional: [
     "performance",
     "performance.getEntriesByType",
+    "window.addEventListener"
```
ender336: awesome thanks

src/errors.ts
ender336: logError is exposed for two reasons - the first is that it enables consumers of clarity to call logerror from their code if they want to log information without throwing exceptions + interfering with execution of the page. 

The second is that it enables testability through phantomjs - phantomjs doesn't support firing custom events but I wanted to trigger the execution in a way the page would actually interact with the method.

gulpfile.ts
```diff
@@ -34,6 +34,28 @@ gulp.task("build", () => {
...
+    "clean",
+    "compile",
+    "place-fixture",
+    "browserify",
+    "uglify",
+    "coverage"
+  );
+});
+
+gulp.task("bnt", () => {
```
ssivov: Can you please a name more descriptive or place a 'decoded' comment ?

clarity.d.ts
```diff
@@ -266,6 +266,9 @@ interface IInstrumentationEventState {
 interface IJsErrorEventState extends IInstrumentationEventState {
   source: string;
   message: string;
+  stack: string;
```
toby-walker: Since you updated the schema please update the schema version (I think it's found in core.ts). Rest looks good.

ender336: sure, I think that Sarvesh is also moving the version number up - let me wait for his to merge then I will change mine.

---
In reply to: [119245274](https://github.com/Microsoft/clarity-js/pull/9#discussion_r119245274) [](ancestors = 119245274)

gulpfile.ts
ender336: I'll add a comment sounds good - I'd prefer to keep it short so I don't have to type too much :P

---
In reply to: [119240679](https://github.com/Microsoft/clarity-js/pull/9#discussion_r119240679) [](ancestors = 119240679)

gulpfile.ts
```diff
@@ -25,6 +25,30 @@ gulp.task("build", () => {
   );
 });
 
+// build and then run coverage
+gulp.task("bnc", () => {
+  runSequence(
+    "clean",
+    "compile",
+    "place-fixture",
+    "browserify",
```
ssivov: There is no more 'browserify'. Please check Sarvesh's changes for a correct merge

ender336: Thanks!

ssivov:
Your branch still has conflicts with the master

ender336:
I think everything is good now - I had also moved to 1.4 so it just auto merged to be a non change :P

---
In reply to: [306346571](https://github.com/Microsoft/clarity-js/pull/9#issuecomment-306346571) [](ancestors = 306346571)

PR #13, opened 2017-06-01
Adding files to package.json

sarveshnagpal:
Required for NPM to include a file from the build folder, in this case: build/clarity.js

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "url": "https://github.com/Microsoft/clarity-js.git"
   },
   "files": [
-    "build/clarity.js"
+    "./build/clarity.js"
   ],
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clarity-js",
-  "version": "0.1.0",
+  "version": "0.1.1",
   "description": "An analytics library that uses web page interactions to generate aggregated insights",
```
toby-walker: Should we start to use the same version for the package and the schema encoded version?

sarveshnagpal: I agree and it makes sense. Are you open to going back to 0.1.1 for the internal version tracking?

PR #14, opened 2017-06-01
Fix package.json

sarveshnagpal:
I misunderstood "files" documentation. Found out .npmignore is a better way.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,11 @@
+# Visual Studio Code temporary files
+.vscode/
+
+# npm packages
+node_modules
+
+# npm logs
+npm-debug.log*
+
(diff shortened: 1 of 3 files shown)


.npmignore
```diff
@@ -0,0 +1,11 @@
...
+.vscode/
+
+# npm packages
+node_modules
+
+# npm logs
+npm-debug.log*
+
+# rollup-plugins-typescript2
+.rpt2_cache
```
ssivov: Does this have to stay in sync with .gitignore ?

sarveshnagpal: Generally yes - will need to check if there's a better way to keep these in sync.

PR #16, opened 2017-06-03
Adding extra JsError safety checks

ssivov:


clarity.d.ts
```diff
@@ -291,6 +291,7 @@ interface ITotalByteLimitExceededEventState extends IInstrumentationEventState {
 
 interface IClarityAssertFailedEventState extends IInstrumentationEventState {
   source: string;
+  comment: string;
```
toby-walker: Bump version. :-)

src/core.ts
```diff
@@ -116,8 +116,8 @@ export function instrument(eventState: IInstrumentationEventState) {
 }
 
 function getUnixTimestamp(): number {
-  return (window.performance && typeof performance.now === "function")
-    ? performance.now() + performance.timing.navigationStart
+  return (window.performance && window.performance.now)
```
toby-walker: Why did you removed the typing check?

src/core.ts
```diff
@@ -135,7 +135,7 @@ function envelope(): IEnvelope {
   return {
     clarityId: cid,
     impressionId,
-    url: top.location.href,
+    url: location.href,
```
toby-walker: Would prefer window.location.href to be clearer.

toby-walker: Why the change from window.top.location to window.location?

toby-walker: Is window.location supported in all browsers? See:

image

From: https://developer.mozilla.org/en-US/docs/Web/API/Window/location

src/plugins/layout.ts
```diff
@@ -68,6 +68,7 @@ export default class Layout implements IPlugin {
     let discoverTime = getTimestamp();
     traverseNodeTree(document, this.discoverNode.bind(this));
     this.shadowDomConsistent = this.shadowDom.mirrorsRealDom();
+    assert(this.shadowDomConsistent, "discoverDom", "shadowDom inconsistent after dom discovery");
```
toby-walker: CAn you return an indication of what inconsistency was? I would encourage you to think about a "diagnostic" approach rather than a simple "pass/fail" approach. You compute diagnostics that indicate issues and can then return what the issues found were and which diagnostics cause a failure. 

This is probably more than the scope of this PR but worth thinking about more.

src/plugins/layout/shadowdom.ts
```diff
@@ -88,7 +92,11 @@ export class ShadowDom {
 
   public updateShadowNode(index: number, newLayout?: ILayoutState) {
     let shadowNode = this.getShadowNode(index);
-    assert(!!shadowNode, "updateShadowNode");
+
+    assert(!!shadowNode, "updateShadowNode", "shadowNode is missing");
```
toby-walker: Add more useful message to indicate what is missing. Add a way to show the element and its location in the tree and include that in the message.

src/plugins/viewport.ts
```diff
@@ -1,8 +1,8 @@
 import { addEvent, bind } from "../core";
 
 export default class Viewport implements IPlugin {
-  private eventName : string = "Viewport";
-  private distanceThreshold : number = 20;
+  private eventName = "Viewport";
+  private distanceThreshold = 20;
```
toby-walker: Low pri but consider moving this to config.

src/core.ts
ssivov: No single strong reason, but:
1. I feel like it's redundant. It can never be anything else but a function (unless manually overriden by someone)
2. I've looked at other projects and how they deal with performance safety checks - vast majority just checks that performance.now exists
3. We've seen errors such as 'object 'performance' has no method 'now'', however every single use of performance.now in our code base was wrapped in this condition. So while i'm not sure how this error is possible in the first place, I decided to tweak a condition and see if it makes a difference.

src/core.ts
ssivov: When the page is loaded in the iframe, 'top.location' will try to access parent page's url, which is forbidden by browser security policies (unless both pages are from the same origin), so it causes a javascript error.

src/plugins/viewport.ts
ssivov: I agree that this has to be in the config as well - we've discussed it on Sarvesh's PR too, but we'll make a separate PR for that.

---
In reply to: [119988354](https://github.com/Microsoft/clarity-js/pull/16#discussion_r119988354) [](ancestors = 119988354)

src/core.ts
ssivov: Here is the more detailed support list. I don't think we should worry about IE < 9 at this point:
image

src/plugins/layout/shadowdom.ts
ssivov: Discussed this offline:
I totally agree that for proper debugging richer and more robust information is needed, so i'll keep working on that, but in a separate PR. For now, we'll go with a plain yes/no check to find the bottleneck of the observed problems.

---
In reply to: [119988311](https://github.com/Microsoft/clarity-js/pull/16#discussion_r119988311) [](ancestors = 119988311)

toby-walker:
:shipit:

PR #17, opened 2017-06-06
Sarveshn/hide images

sarveshnagpal:
- Add a configuration option to mask images
- Also mask text present in HTML attributes like: value, placeholder, title or alt

### DIFF
--- a/test/layout.ts
+++ b/test/layout.ts
@@ -622,4 +622,36 @@ describe("Layout Tests", () => {
       done();
     }
   });
+
+  it("checks that input value is masked if the config is set to not show text", (done) => {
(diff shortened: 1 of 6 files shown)


test/layout.ts
```diff
@@ -556,4 +556,70 @@ describe("Layout Tests", () => {
       callbackNumber++;
     }
   });
+
+  it("checks that images source is not captured if the config disallows it", (done) => {
```
toby-walker: Add some test to cover masking of the attributes you added in addition to the src field.

sarveshnagpal: Good point, done.

PR #20, opened 2017-06-27
Removing JSErrors

ssivov:


src/core.ts
```diff
@@ -4,7 +4,7 @@ import getPlugin from "./plugins";
 import { debug, getCookie, guid, isNumber, mapProperties, setCookie } from "./utils";
 
 // Constants
-const version = "0.1.7";
+const version = "0.1.8";
```
toby-walker: Minor: Duplicates number in package.json. Any way to get them from one place?

src/core.ts
```diff
@@ -117,8 +117,8 @@ export function instrument(eventState: IInstrumentationEventState) {
 }
 
 function getUnixTimestamp(): number {
-  return (window.performance && window.performance.now)
```
toby-walker: Should you not check for the presence of windows.timing as well?

src/plugins/layout.ts
```diff
@@ -96,40 +96,42 @@ export default class Layout implements IPlugin {
...
-            case Node.TEXT_NODE:
-              (layoutState as ITextLayoutState).content = originalProperties.characterData;
-              break;
-            case Node.ELEMENT_NODE:
-              let originalAttributes = (layoutState as IElementLayoutState).attributes;
-              for (let attr in originalAttributes) {
-                if (originalAttributes.hasOwnProperty(attr)) {
-                  (layoutState as IElementLayoutState).attributes[attr] = originalAttributes[attr];
+
+      if (shadowNode) {
```
toby-walker: When can this be null?

src/utils.ts
```diff
@@ -6,8 +6,9 @@ import { instrument } from "./core";
 // tslint:disable
 export function guid() {
   let d = new Date().getTime();
-  if (window.performance && window.performance.now) {
```
toby-walker: Minor. You might want to put this in a function as it's duplicated in several places.

src/plugins/layout.ts
ssivov: In design - never, however it is sometimes. I'm on a mission to find out when this happens as my next task.

src/core.ts
ssivov: There are some cases when performance object exists, without the 'now' method, but it always has the 'timing' property

src/core.ts
ssivov: Unfortunately, no

src/plugins/layout.ts
ender336: We should explicitly log something in an else clause then so we have a way to investigate further

---
In reply to: [124159155](https://github.com/Microsoft/clarity-js/pull/20#discussion_r124159155) [](ancestors = 124159155)

PR #21, opened 2017-07-14
Create plugin to allow Clarity to track custom site-specific data using DOM events

james-schend-ms:
Allows sites with the need to log data specific to themselves to send an DOM "claritydata" event; the data will be logged by Clarity where it can be joined to other behavioral data collected from the page.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -10,5 +10,5 @@ export let config: IConfig = {
   instrument: false,
   uploadHandler: null,
   debug: false,
-  plugins: ["viewport", "layout", "pointer", "performance", "errors"]
+  plugins: ["viewport", "layout", "pointer", "performance", "errors", "customdata"]
 };
--- a/src/plugins/customdata.ts
+++ b/src/plugins/customdata.ts
@@ -1,11 +1,7 @@
 // Allows data sent via. normal DOM event handlers to be logged alongside Clarity data by listening to the
 // custom DOM event "claritydata".
+// Custom event documentation: https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent
 import { addEvent, bind } from "./../core";
-
-// Events coming from normal DOM will have additional data in details property
-interface ICustomEvent extends Event {
-    detail: any;
-}
 
 export default class CustomData implements IPlugin {
     private eventName = "Custom";
@@ -23,6 +19,6 @@ export default class CustomData implements IPlugin {
     }
 
     private customDataHandler(customEvent: ICustomEvent) {
-        addEvent(this.eventName, { detail: customEvent.detail });
+		addEvent(this.eventName, { type: customEvent.detail.type, data: customEvent.detail.data });
     }
 }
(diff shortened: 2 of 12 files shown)


ssivov:
Please bump the minor version in core.ts and package.json

james-schend-ms:
Bumped the minor version number.

src/config.ts
```diff
@@ -10,5 +10,5 @@ export let config: IConfig = {
   instrument: false,
   uploadHandler: null,
   debug: false,
-  plugins: ["viewport", "layout", "pointer", "performance", "errors"]
+  plugins: ["viewport", "layout", "pointer", "performance", "errors", "customdata"]
```
toby-walker: Could you add a MD (markdown) file about to use the add-on with an sample/example?

src/plugins/customdata.ts
```diff
@@ -0,0 +1,28 @@
+// Allows data sent via. normal DOM event handlers to be logged alongside Clarity data by listening to the
+// custom DOM event "claritydata".
+import { addEvent, bind } from "./../core";
+
+// Events coming from normal DOM will have additional data in details property
+interface ICustomEvent extends Event {
+    detail: any;
```
toby-walker: The typing here is weak. Is there any data contract at all? How would I distinguish different kinds of custom events? At east indicate the type of event.

src/plugins/customdata.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+export default class CustomData implements IPlugin {
+    private eventName = "Custom";
+
+    public activate() {
+        bind(window, "claritydata", this.customDataHandler.bind(this));
```
toby-walker: Is the idea that the user sends claritydata events? See other comment about need for brief documentation about how this module should be used.

src/plugins/customdata.ts
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    public reset(): void {
+        return;
+    }
+
+    public teardown() {
```
toby-walker: unbind the handler.

src/config.ts
james-schend-ms: Added. I just put it inside the plugins directory, I hope that is the right location.

src/plugins/customdata.ts
james-schend-ms: Had a quick conversation with Toby 7/21. Since current Clarity modules (including pointer and errors) do not unbind handlers on cleanup, this behavior is consistent. Should consider a future conversation about whether Clarity is ever expected to be unloaded independent of page unload.

src/plugins/customdata.ts
james-schend-ms: Created a new type to specify that Clarity requires a "type" and "data" for each custom event log. Requiring a "type" should make data easier to query for analysts.

src/plugins/customdata.ts
james-schend-ms: Added a comment at the top of the file to the DOM documentation for .createEvent() which should help. Also see documentation added in customdata.md.

james-schend-ms:
Updated to address all outstanding comments. Thanks for the review.

PR #24, opened 2017-07-28
Log page title in the envelope

sarveshnagpal:
Minor change to log page title along with rest of the fields in the envelope. Helps with visualization.

### DIFF
--- a/clarity.d.ts
+++ b/clarity.d.ts
@@ -58,6 +58,7 @@ interface IEnvelope {
   clarityId: string;
   impressionId: string;
   url: string;
+  title: string;
   version: string;
   time: number;
   sequenceNumber: number;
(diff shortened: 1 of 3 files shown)


clarity.d.ts
```diff
@@ -58,6 +58,7 @@ interface IEnvelope {
   clarityId: string;
   impressionId: string;
   url: string;
+  title: string;
```
toby-walker: Sending this in every envelope seems consistent with the current design but a not great design. Maybe it's time to change this and create an explicit event with "page data" that includes this and send it and other common information only once rather than as part of every envelope? Also do we not pick up changes to the <title> element using normal DOM mutations?

sarveshnagpal: You're right and let's hold off on this change until we fix the overall design to not send same data multiple times. Url, title, etc. all add up. Thanks for reviewing.

PR #25, opened 2017-08-01
A bundle of changes to improve data quality and consistency

ssivov:
* Removed superfluous assertions 
* Improving ShadowDOM consistency checking
* Created instrumentation event for ShadowDOM inconsistency
* Added last consistent dom state to instrumentation
* Make layout process mutations and apply them on the ShadowDom immediately
* Fixed bugs in backfilling node's initial state and made a change to send layout states of ignored nodes
* Add characterDataOldValue to the MutationObserver
* Detached layout state storage from ShadowDom + Bug Fixes
* Fixed bug where some removed nodes maintained their index
* Added new Layout test and improved some existing ones
* Removed more assertions and obsolete 'Ignore' action
* Removed 'original properties' feature
* Improving checks for nodes that should be ignored + Tests
* Removing index gaps between new nodes + Tests
* Removing oldValue MutationObserver properties
* Moved 'ignore' flag setting to insertShadowNode function

### DIFF
--- a/clarity.d.ts
+++ b/clarity.d.ts
@@ -35,6 +35,9 @@ interface IConfig {
   // Setting to enable debug features (e.g. console.log statements)
   debug?: boolean;
 
+  // Setting to enable consistency verifications between real DOM and shadow DOM 
+  ensureConsistency?: boolean;
+
   // Active plugins
   plugins?: string[];
 }
--- a/clarity.d.ts
+++ b/clarity.d.ts
@@ -99,7 +99,6 @@ type UploadHandler = (payload: string, onSuccess?: UploadCallback, onFailure?: U
 
 interface IShadowDomNode extends HTMLDivElement {
   node: Node; /* Reference to the node in the real DOM */
-  layout: ILayoutState; /* Latest layout JSON */
   ignore: boolean;  /* Flag to avoid sending data for that node */
 }
 
--- a/clarity.d.ts
+++ b/clarity.d.ts
@@ -309,10 +309,11 @@ interface IClarityDuplicatedEventState extends IInstrumentationEventState {
 }
 
 interface IShadowDomInconsistentEventState extends IInstrumentationEventState {
-  dom: string;
-  shadowDom: string;
+  dom: object;
+  shadowDom: object;
+  lastConsistentShadowDom: object;
   lastAction: string;
-  lastConsistentShadowDom?: string;
+  firstEvent?: IShadowDomInconsistentEventState;
 }
 
 /* ##################################### */
--- a/clarity.d.ts
+++ b/clarity.d.ts
@@ -167,6 +167,10 @@ interface ITextLayoutState extends ILayoutState {
   content: string;
 }
 
+interface IIgnoreLayoutState extends ILayoutState {
+  nodeType: number;
+}
+
 interface IMutationEntry {
   node: Node;
   action: Action;
--- a/clarity.d.ts
+++ b/clarity.d.ts
@@ -36,7 +36,11 @@ interface IConfig {
   debug?: boolean;
 
   // Setting to enable consistency verifications between real DOM and shadow DOM
-  ensureConsistency?: boolean;
+  // Validating consistency can be costly performance-wise, because it requires
+  // re-traversing entire DOM and ShadowDom to compare them against each other.
+  // The upside is knowing deterministically that all activity on the page was
+  // interpreted correctly and data is reliable.
+  validateConsistency?: boolean;
 
   // Active plugins
   plugins?: string[];
@@ -187,6 +191,20 @@ interface IMutationEntry {
   next?: Node;
 }
 
+declare const enum LayoutRoutine {
+  DiscoverDom,
+  Mutation
+}
+
+interface ILayoutRoutineInfo {
+  action: LayoutRoutine;
+}
+
+interface IMutationRoutineInfo extends ILayoutRoutineInfo {
+  mutationSequence: number; /* Sequence number of the mutation batch */
+  batchSize: number;  /* Number of mutation records in the mutation callback */
+}
+
 // Interface to store some information about the initial state of the node
 // in cases where mutation happens before we process the node for the first time.
 // Stroring original values of the properties that can were mutated allows us to
@@ -324,7 +342,7 @@ interface IShadowDomInconsistentEventState extends IInstrumentationEventState {
   lastConsistentShadowDom: NumberJson;
 
   // Last action that happened before we found out that ShadowDom is inconsistent
-  lastAction: string;
+  lastAction: ILayoutRoutineInfo;
 
   // To handle specific MutationObserver behavior in IE, we wait for ShadowDom to become inconsistent twice in a row,
   // before we stop processing mutations and send ShadowDomInconsistentEvent. This means that the actual transition
--- a/clarity.d.ts
+++ b/clarity.d.ts
@@ -122,8 +122,7 @@ declare const enum Action {
   Insert,
   Update,
   Remove,
-  Move,
-  Ignore
+  Move
 }
 
 interface IAttributes {
--- a/clarity.d.ts
+++ b/clarity.d.ts
@@ -262,7 +262,8 @@ declare const enum Instrumentation {
   TotalByteLimitExceeded,
   Teardown,
   ClarityAssertFailed,
-  ClarityDuplicated
+  ClarityDuplicated,
+  ShadowDomInconsistent
 }
 
 interface IInstrumentationEventState {
@@ -303,6 +304,14 @@ interface IClarityAssertFailedEventState extends IInstrumentationEventState {
 interface IClarityDuplicatedEventState extends IInstrumentationEventState {
   currentImpressionId: string;
 }
+
+interface IShadowDomInconsistentEventState extends IInstrumentationEventState {
+  dom: string;
+  shadowDom: string;
+  lastAction: string;
+  lastConsistentShadowDom?: string;
+}
+
 /* ##################################### */
 /* #########   PERFORMANCE   ########### */
 /* ##################################### */
--- a/clarity.d.ts
+++ b/clarity.d.ts
@@ -135,6 +135,7 @@ interface IAttributes {
 // different layout events originating from different actions
 interface ILayoutEventInfo {
   node: Node;
+  index: number;
   source: Source;
   action: Action;
   time?: number;
--- a/clarity.d.ts
+++ b/clarity.d.ts
@@ -100,6 +100,10 @@ type UploadHandler = (payload: string, onSuccess?: UploadCallback, onFailure?: U
 /* ############   LAYOUT   ############# */
 /* ##################################### */
 
+type NumberJson = {
+  [key: number]: NumberJson;
+};
+
 interface IShadowDomNode extends HTMLDivElement {
   node: Node; /* Reference to the node in the real DOM */
   ignore: boolean;  /* Flag to avoid sending data for that node */
@@ -172,6 +176,7 @@ interface ITextLayoutState extends ILayoutState {
 
 interface IIgnoreLayoutState extends ILayoutState {
   nodeType: number;
+  elementTag?: string;
 }
 
 interface IMutationEntry {
@@ -309,10 +314,22 @@ interface IClarityDuplicatedEventState extends IInstrumentationEventState {
 }
 
 interface IShadowDomInconsistentEventState extends IInstrumentationEventState {
-  dom: object;
-  shadowDom: object;
-  lastConsistentShadowDom: object;
+  // JSON of node indicies, representing the DOM
+  dom: NumberJson;
+
+  // JSON of ShadowNode IDs, representing the inconsistent ShadowDom
+  shadowDom: NumberJson;
+
+  // JSON of ShadowNode IDs, representing the last consistent ShadowDom
+  lastConsistentShadowDom: NumberJson;
+
+  // Last action that happened before we found out that ShadowDom is inconsistent
   lastAction: string;
+
+  // To handle specific MutationObserver behavior in IE, we wait for ShadowDom to become inconsistent twice in a row,
+  // before we stop processing mutations and send ShadowDomInconsistentEvent. This means that the actual transition
+  // from consistent to inconsistent state happened on some previous action and there was also an event created for it.
+  // That first event is sent in this property.
   firstEvent?: IShadowDomInconsistentEventState;
 }
 
--- a/clarity.d.ts
+++ b/clarity.d.ts
@@ -141,18 +141,15 @@ interface ILayoutEventInfo {
   time?: number;
 }
 
-interface ILayoutStateBase {
+interface ILayoutState {
   index: number;  /* Index of the layout element */
+  tag: string;  /* Tag name of the element */
   source: Source; /* Source of discovery */
   action: Action; /* Reflect the action with respect to DOM */
-  mutationSequence?: number;  /* Sequence number of the mutation batch */
-}
-
-interface ILayoutState extends ILayoutStateBase {
   parent: number; /* Index of the parent element */
-  tag: string;  /* Tag name of the element */
   previous: number; /* Index of the previous sibling, if known */
   next: number; /* Index of the next sibling, if known */
+  mutationSequence?: number;  /* Sequence number of the mutation batch */
 }
 
 interface IDoctypeLayoutState extends ILayoutState {
--- a/clarity.d.ts
+++ b/clarity.d.ts
@@ -35,7 +35,7 @@ interface IConfig {
   // Setting to enable debug features (e.g. console.log statements)
   debug?: boolean;
 
-  // Setting to enable consistency verifications between real DOM and shadow DOM 
+  // Setting to enable consistency verifications between real DOM and shadow DOM
   ensureConsistency?: boolean;
 
   // Active plugins
--- a/src/plugins/layout.ts
+++ b/src/plugins/layout.ts
@@ -81,8 +81,7 @@ export default class Layout implements IPlugin {
 
   // Add node to the ShadowDom to store initial adjacent node info in a layout and obtain an index
   private discoverNode(node: Node) {
-    let ignore = shouldIgnoreNode(node, this.shadowDom);
-    this.shadowDom.insertShadowNode(node, getNodeIndex(node.parentNode), getNodeIndex(node.nextSibling), ignore);
+    this.shadowDom.insertShadowNode(node, getNodeIndex(node.parentNode), getNodeIndex(node.nextSibling));
     let index = getNodeIndex(node);
     let layout = createGenericLayoutState(node, null);
     this.layoutStates[index] = layout;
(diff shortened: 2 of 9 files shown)


clarity.d.ts
```diff
@@ -303,6 +304,14 @@ interface IClarityAssertFailedEventState extends IInstrumentationEventState {
 interface IClarityDuplicatedEventState extends IInstrumentationEventState {
   currentImpressionId: string;
 }
+
+interface IShadowDomInconsistentEventState extends IInstrumentationEventState {
```
toby-walker: what are these value of the properties of these events? Can you add more comments about what this event is recording?

src/plugins/layout/shadowdom.ts
```diff
@@ -1,10 +1,9 @@
 import { assert, isNumber, traverseNodeTree } from "../../utils";
-import { getNodeIndex, IgnoreTag, NodeIndex } from "./stateprovider";
+import { getNodeIndex, NodeIndex, shouldIgnoreNode } from "./stateprovider";
 
 // Class names for child list mutation classifications
 const FinalClassName = "cl-final";
 const NewNodeClassName = "cl-new";
-const TopNewNodeClassName = "cl-new-top";
 const MovedNodeClassName = "cl-moved";
 const UpdatedNodeClassName = "cl-updated";
```
toby-walker: Based on the code below it seemed like "Moved" represented "Deleted". If that's correct, rename to reflect that.

src/plugins/layout/shadowdom.ts
```diff
@@ -26,39 +25,33 @@ export class ShadowDom {
...
-      if (!this.hasClass(parent, NewNodeClassName)) {
-        this.setClass(shadowNode, TopNewNodeClassName);
-      }
     }
 
-    if (nextSibling) {
-      parent.insertBefore(shadowNode, nextSibling);
-    } else {
-      parent.appendChild(shadowNode);
+    assert(!!parent, "insertShadowNode", "parent is missing");
```
toby-walker: is this truly an erroneous situation?

src/plugins/layout/shadowdom.ts
```diff
@@ -69,59 +62,39 @@ export class ShadowDom {
...
-      if (newLayout) {
-        shadowNode.layout = newLayout;
-        shadowNode.ignore = (newLayout.tag === IgnoreTag);
-      }
-
-      if (this.classifyNodes) {
-        this.setClass(shadowNode, UpdatedNodeClassName);
-      }
+    if (shadowNode && this.classifyNodes) {
+      this.setClass(shadowNode, UpdatedNodeClassName);
```
toby-walker: CAn you describe the states that these are encoding, what each means and what the possible transitions between the states are? A comment that did this would be very helpful for understanding the code.

src/plugins/layout/shadowdom.ts
```diff
@@ -240,22 +187,11 @@ export class ShadowDom {
...
-          this.removeClass(next, NewNodeClassName);
-          // Add children right-to-left
-          let children = next.childNodes;
-          for (let i = children.length - 1; i >= 0; i--) {
-            discoverQueue.push(children[i] as IShadowDomNode);
-          }
-        }
-      }
+    let newNodes = this.doc.getElementsByClassName(NewNodeClassName);
+    while (newNodes.length > 0) {
```
toby-walker: This like like an infinite loop. Where do you shrink the list of newNodes?

src/plugins/layout/shadowdom.ts
```diff
@@ -272,25 +208,93 @@ export class ShadowDom {
...
+        }
+      });
     }
 
     return summary;
   }
 
+  public createDomIndexJson(): object {
+
+    function writeIndex(node: Node, json: object) {
```
toby-walker: can't you find a better type for this? Something to indicate it is a mapping from numbers to maps of numbers?

src/plugins/layout/shadowdom.ts
```diff
@@ -272,25 +208,93 @@ export class ShadowDom {
...
+      }
+    }
+
+    let indexJson = {};
+    writeIndex(document, indexJson);
+
+    return indexJson;
+  }
+
+  public createShadowDomIndexJson(): object {
```
toby-walker: Code looks nearly identical to above. Can you remove the duplication by refactoring?

src/plugins/layout/shadowdom.ts
```diff
@@ -272,25 +208,93 @@ export class ShadowDom {
...
+    writeIndex(this.shadowDocument, indexJson);
+
+    return indexJson;
+  }
+
+  public isConsistent(): boolean {
+    return this.validateNodeWithSubtree(document, this.shadowDocument);
+  }
+
+  private validateShadow(node: Node, shadowNode: IShadowDomNode) {
```
toby-walker: 1. Rename to "isConsistentNode"
2. Should indicate it returns a boolean.

src/plugins/layout/shadowdom.ts
```diff
@@ -272,25 +208,93 @@ export class ShadowDom {
...
+    let index = getNodeIndex(node);
+    return (isNumber(index) && shadowNode.id === (index).toString() && shadowNode.node === node);
+  }
+
+  private validateNodeWithSubtree(node: Node, shadowNode: IShadowDomNode): boolean {
```
toby-walker: Rename to isConstentSubtree as it return a value.

src/plugins/layout/shadowdom.ts
```diff
@@ -347,6 +350,22 @@ export class ShadowDom {
...
 
+  // When we apply a mutation batch, we assign a new index to every node that is added to the DOM as we parse mutations
+  // in their natural order. However, some of those nodes end up being again removed from DOM with the some follow-up mutation
+  // thus 'stealing' an index with them. Since we don't instrument nodes that were added and removed within the same
+  // mutation batch, we have no records of these stolen indices. To maintain consistency on the backend, at the end of the mutation,
+  // we can re-assign indices for all nodes that remained attached to the DOM such that they all go in increasing order without gaps
+  private reIndexNewNodes(nextIndex: number) {
+    let newNodes = this.doc.getElementsByClassName(NewNodeClassName);
+    for (let i = 0; i < newNodes.length; i++) {
+      let shadowNode = newNodes[i] as IShadowDomNode;
```
toby-walker: Minor: Type the newNodes = array<IShadowDomNode> instead.

src/plugins/layout/stateprovider.ts
```diff
@@ -117,6 +106,12 @@ export function createTextLayoutState(textNode: Text): ITextLayoutState {
   return textState;
 }
 
+export function createIgnoreLayoutState(node: Node): IIgnoreLayoutState {
```
toby-walker: Can we include the type of node that was ignored? Useful to know.

src/plugins/layout/stateprovider.ts
```diff
@@ -129,3 +124,34 @@ export function createGenericLayoutState(node: Node, tag: string): ILayoutState
...
   return layoutState;
 }
+
+export function shouldIgnoreNode(node: Node, shadowDom: ShadowDom): boolean {
+  let shadowNode = shadowDom.getShadowNode(getNodeIndex(node));
+  let ignore = false;
+  switch (node.nodeType) {
+    case Node.ELEMENT_NODE:
+      let tagName = (node as Element).tagName;
+      if (tagName === "SCRIPT" || tagName === "META") {
```
toby-walker: 1. Make the set of nodes we ignore configurable. Default to these.
2. Why are we not sending these back now? At least he <meta> seems useful to have. 
3. It would be useful to send back script tags but not the children. Could we do that?

src/plugins/layout/stateprovider.ts
```diff
@@ -129,3 +124,34 @@ export function createGenericLayoutState(node: Node, tag: string): ILayoutState
...
+  }
+
+  // Ignore subtrees of ignored nodes (e.g. text with a <script> parent)
+  if (!ignore) {
+    let parentIndex = getNodeIndex(node.parentNode);
+    if (parentIndex !== null) {
+      let parentShadowNode = shadowDom.getShadowNode(parentIndex);
+      assert(!!parentShadowNode, "shouldIgnoreNode", "parentShadowNode is missing");
+      if (parentShadowNode && parentShadowNode.ignore && parentShadowNode.node !== document) {
+        ignore = true;
```
toby-walker: Can break here.

src/plugins/layout/shadowdom.ts
ssivov: This can't be done, because newNodes is an HTMLCollection, which cannot be converted to an array.

---
In reply to: [131240755](https://github.com/Microsoft/clarity-js/pull/25#discussion_r131240755) [](ancestors = 131240755)

src/plugins/layout/stateprovider.ts
ssivov: There is no loop to break

---
In reply to: [131241578](https://github.com/Microsoft/clarity-js/pull/25#discussion_r131241578) [](ancestors = 131241578)

src/plugins/layout/stateprovider.ts
ssivov: Yeah, we can do that, but i think it deserves a separate sync on what kind of nodes are we willing to ignore and which not, as well as a separate PR.

---
In reply to: [131241375](https://github.com/Microsoft/clarity-js/pull/25#discussion_r131241375) [](ancestors = 131241375)

src/plugins/layout/stateprovider.ts
ssivov: I agree! Node type is already recorded 2 lines below. However, if ignored node was an Element, i will also add it's original tagName to the state (e.g. 'SCRIPT')

---
In reply to: [131241078](https://github.com/Microsoft/clarity-js/pull/25#discussion_r131241078) [](ancestors = 131241078)

src/plugins/layout/shadowdom.ts
ssivov: newNodes is a 'live' HTMLCollection of nodes with the requested className. When I remove class from a node, that node automatically disappears from that collection.

---
In reply to: [131238745](https://github.com/Microsoft/clarity-js/pull/25#discussion_r131238745) [](ancestors = 131238745)

src/plugins/layout/shadowdom.ts
ssivov: Created NumberJson type, however the compiler doesn't seem to enforce numeric keys. Still looks better for code readability though, thanks.

---
In reply to: [131239582](https://github.com/Microsoft/clarity-js/pull/25#discussion_r131239582) [](ancestors = 131239582)

src/plugins/layout/shadowdom.ts
ssivov: Done, merged two functions in one with a help of an extra argument

---
In reply to: [131239745](https://github.com/Microsoft/clarity-js/pull/25#discussion_r131239745) [](ancestors = 131239745)

src/plugins/layout/shadowdom.ts
ssivov: Yes, when we insert a node, we always attached it to some already existing node. Even if the parent was removed from DOM in some previous mutation, its shadow node would still be in the ShadowDOM's special 'removed' container, until we process all the mutations, so it should be accessible until the very end.

---
In reply to: [131237635](https://github.com/Microsoft/clarity-js/pull/25#discussion_r131237635) [](ancestors = 131237635)

clarity.d.ts
ssivov: Added comments

---
In reply to: [131236965](https://github.com/Microsoft/clarity-js/pull/25#discussion_r131236965) [](ancestors = 131236965)

src/plugins/layout/shadowdom.ts
ssivov: Added comments to 'applyMutationBatch' and 'getMutationSummary' functions, explaining in detail what's going on there.

---
In reply to: [131238248](https://github.com/Microsoft/clarity-js/pull/25#discussion_r131238248) [](ancestors = 131238248)

src/plugins/layout/shadowdom.ts
ssivov: Remove action is processed as moving the node to the 'removedNodes' container. Since every 'Move' is really a remove followed by an insert instantly, we can only deterministically say that node was really removed, when we process all mutations and node ends up in the 'removedNodes' container. Please see the comments I added before the functions for richer details.

---
In reply to: [131237174](https://github.com/Microsoft/clarity-js/pull/25#discussion_r131237174) [](ancestors = 131237174)

src/plugins/layout.ts
```diff
@@ -411,5 +334,38 @@ export default class Layout implements IPlugin {
       });
     }
 
+    return events;
+  }
+
+  private ensureConsistency(lastAction: string): void {
+    if (config.ensureConsistency) {
```
ender336: >ensureConsistency [](start = 15, length = 17)

I think we should switch either the flag name or the method name to disambiguate them

src/plugins/layout/shadowdom.ts
```diff
@@ -69,59 +63,55 @@ export class ShadowDom {
 
     assert(!!parent, "moveShadowNode", "parent is missing");
```
ender336: >assert [](start = 4, length = 6)

maybe out of scope for here, but should we turn off clarity once some threshold of assertions fail?

src/plugins/layout/shadowdom.ts
```diff
@@ -240,29 +214,18 @@ export class ShadowDom {
...
-          summary.newNodes.push(next);
-          this.removeClass(next, NewNodeClassName);
-          // Add children right-to-left
-          let children = next.childNodes;
-          for (let i = children.length - 1; i >= 0; i--) {
-            discoverQueue.push(children[i] as IShadowDomNode);
-          }
-        }
-      }
+    let newNodes = this.doc.getElementsByClassName(NewNodeClassName);
```
ender336: >getElementsByClassName [](start = 28, length = 22)

since we're in here, maybe add TODOs on the getElementsByCLassName indicating we need to get off of Live HTMLCollections

src/plugins/layout/stateprovider.ts
```diff
@@ -129,3 +127,34 @@ export function createGenericLayoutState(node: Node, tag: string): ILayoutState
...
   return layoutState;
 }
+
+export function shouldIgnoreNode(node: Node, shadowDom: ShadowDom): boolean {
+  let shadowNode = shadowDom.getShadowNode(getNodeIndex(node));
+  let ignore = false;
+  switch (node.nodeType) {
+    case Node.ELEMENT_NODE:
+      let tagName = (node as Element).tagName;
+      if (tagName === "SCRIPT" || tagName === "META") {
```
ender336: >"META" [](start = 46, length = 6)

a bit of a nit, but for these string constants that get used multiple times, you'll get better crunching if you move them to a variable

ender336:
:shipit:

src/plugins/layout/shadowdom.ts
ssivov: I think it's ok to keep them, because they don't necessarily cause critical issues, but rather lets us know that there is something wrong with the logic. So, if assertion fails, but DOM still arrives to consistent state, then we can continue collecting data safely. Otherwise, we'll arrive to the inconsistent state and stop processing mutations

---
In reply to: [132226650](https://github.com/Microsoft/clarity-js/pull/25#discussion_r132226650) [](ancestors = 132226650)

src/plugins/layout/shadowdom.ts
ssivov: Definitely a big work item! We tried to avoid in-code TODOs, but we can create a tracking item on github

---
In reply to: [132227437](https://github.com/Microsoft/clarity-js/pull/25#discussion_r132227437) [](ancestors = 132227437)

src/plugins/layout/stateprovider.ts
ssivov: Moved to a const like the other tags up top

---
In reply to: [132228127](https://github.com/Microsoft/clarity-js/pull/25#discussion_r132228127) [](ancestors = 132228127)

ender336:
test comment

src/plugins/layout.ts
ssivov: Changed method name to 'checkConsistency'

---
In reply to: [132224906](https://github.com/Microsoft/clarity-js/pull/25#discussion_r132224906) [](ancestors = 132224906)

clarity.d.ts
```diff
@@ -35,6 +35,9 @@ interface IConfig {
   // Setting to enable debug features (e.g. console.log statements)
   debug?: boolean;
 
+  // Setting to enable consistency verifications between real DOM and shadow DOM
+  ensureConsistency?: boolean;
```
sarveshnagpal: >ensureConsistency?: boolean; [](start = 2, length = 28)

validateConsistency? By default we will try to ensure consistency but I believe you are trying here to validate that we were indeed consistent? Also a quick comment on potential cost would be nice for folks to understand the associated cost of the feature.

clarity.d.ts
```diff
@@ -303,6 +312,27 @@ interface IClarityAssertFailedEventState extends IInstrumentationEventState {
 interface IClarityDuplicatedEventState extends IInstrumentationEventState {
   currentImpressionId: string;
 }
+
+interface IShadowDomInconsistentEventState extends IInstrumentationEventState {
+  // JSON of node indicies, representing the DOM
+  dom: NumberJson;
```
sarveshnagpal: >NumberJson [](start = 7, length = 10)

Nitpick: Did you mean to send back JSON specific to a node? If so, would NodeJson or something similar be a more descriptive name?

src/plugins/layout.ts
```diff
@@ -67,18 +75,22 @@ export default class Layout implements IPlugin {
   private discoverDom() {
     let discoverTime = getTimestamp();
     traverseNodeTree(document, this.discoverNode.bind(this));
-    this.shadowDomConsistent = this.shadowDom.mirrorsRealDom();
-    assert(this.shadowDomConsistent, "discoverDom", "shadowDom inconsistent after dom discovery");
-    this.backfillLayoutsAsync(discoverTime, this.onDomDiscoverComplete.bind(this));
+    this.checkConsistency("Discover DOM");
```
sarveshnagpal: >"Discover DOM" [](start = 26, length = 14)

use enum instead?

sarveshnagpal:
:shipit:

clarity.d.ts
ssivov: Sounds like a better name, yeah - made the change


---
In reply to: [132250973](https://github.com/Microsoft/clarity-js/pull/25#discussion_r132250973) [](ancestors = 132250973)

clarity.d.ts
ssivov: In a way..sending back the entire document here, so it is specific to a node - document node :) However, since we represent nodes with numeric indices it's ends up being JSON with numeric keys, hence the new type.

---
In reply to: [132251527](https://github.com/Microsoft/clarity-js/pull/25#discussion_r132251527) [](ancestors = 132251527)

src/plugins/layout.ts
ssivov: Switched to using enums and added some more structure to the mutation inconsistency event (mutation sequence # and number of mutation records)

---
In reply to: [132251947](https://github.com/Microsoft/clarity-js/pull/25#discussion_r132251947) [](ancestors = 132251947)

PR #27, opened 2017-08-11
Performance optimizations

ssivov:


src/core.ts
```diff
@@ -93,26 +92,32 @@ export function addEvent(type: string, eventState: any, time?: number) {
...
-      clearTimeout(timeout);
-    }
+  scheduleUpload = scheduleUpload && !(event.state && event.state.type === Instrumentation.XhrError && nextPayload.length === 1);
+  if (scheduleUpload) {
+    clearTimeout(timeout);
     timeout = setTimeout(uploadNextPayload, config.delay);
   }
 }
 
+export function addMultipleEvents(events: IEventData[]) {
```
sarveshnagpal: >addMultipleEvents [](start = 16, length = 17)

Where is this used?

src/plugins/layout/shadowdom.ts
```diff
@@ -365,8 +367,11 @@ export class ShadowDom {
     let newNodes = this.doc.getElementsByClassName(NewNodeClassName);
     for (let i = 0; i < newNodes.length; i++) {
       let shadowNode = newNodes[i] as IShadowDomNode;
+      let currentIndex = getNodeIndex(shadowNode);
       shadowNode.id = "" + nextIndex;
       shadowNode.node[NodeIndex] = nextIndex;
+      delete this.nodes[currentIndex];
```
sarveshnagpal: >delete this.nodes[currentIndex]; [](start = 6, length = 32)

Generally, delete is a slow operation. Can we just leave it as undefined?

src/core.ts
ssivov: Added another commit to process all events in batches, when we have many at the same time, instead of one by one

src/plugins/layout/shadowdom.ts
ssivov: Ran some tests and looks like it is, indeed, almost twice faster. The only issue is that setting some property (or array index) to undefined, still counts as having that property in the object/array with value 'undefined'. So, if we do this:

var a = [];
a[5] = 5;
a[5] = undefined;

Then, a.length === 6. Don't think we're using nodes.length anywhere right now, but this could be a bit dangerous if someone later on decides to query the number of observed nodes through nodes.length.

src/core.ts
```diff
@@ -93,26 +92,32 @@ export function addEvent(type: string, eventState: any, time?: number) {
...
-  // Don't reschedule upload when next payload consists of exactly one XhrError instrumentation event.
+  // Don't schedule next upload when next payload consists of exactly one XhrError instrumentation event.
   // This helps us avoid the infinite loop in the case when all requests fail (e.g. dropped internet connection)
   // Infinite loop comes from sending instrumentation about failing to deliver previous delivery failure instrumentation.
-  let rescheduleUpload = !(eventState && eventState.type === Instrumentation.XhrError && nextPayload.length === 1);
-  if (rescheduleUpload) {
-    if (timeout) {
-      clearTimeout(timeout);
-    }
+  scheduleUpload = scheduleUpload && !(event.state && event.state.type === Instrumentation.XhrError && nextPayload.length === 1);
```
ender336: >scheduleUpload [](start = 2, length = 14)

nit: I'm against modifying input parameters like this as a practice and I don't think the byte count savings are worth the readability/maintainability.

src/core.ts
```diff
@@ -93,26 +92,32 @@ export function addEvent(type: string, eventState: any, time?: number) {
...
+    clearTimeout(timeout);
     timeout = setTimeout(uploadNextPayload, config.delay);
   }
 }
 
+export function addMultipleEvents(events: IEventData[]) {
+  for (let i = 0; i < events.length; i++) {
+    // Only schedule upload after the last event
+    let scheduleUpload = (i === events.length - 1);
+    addEvent(events[i], scheduleUpload);
```
ender336: probably a really minor optimization, but we always know we want scheduleUpload to be true on the final iteration so we should just loop until i < events.length - 1 and always pass false within the loop, then make one more explicit call after with events[events.length-1] and true

src/plugins/layout/shadowdom.ts
```diff
@@ -15,15 +15,16 @@ export class ShadowDom {
   private removedNodes = this.doc.createElement("div");
   private shadowDomRoot = this.doc.createElement("div");
   private classifyNodes = false;
+  private nodes: IShadowDomNode[] = [];
```
ender336: >nodes [](start = 10, length = 5)

nit: rename to something like indexToNodeMap? I tend to err on the side of too expressive, so maybe meet somewhere in the middle :P

src/plugins/layout/shadowdom.ts
ender336: i think since nodes is just a lookup table we will be ok even long term. 

---
In reply to: [133095363](https://github.com/Microsoft/clarity-js/pull/27#discussion_r133095363) [](ancestors = 133095363)

src/core.ts
ssivov: Small, but nice change, thanks!

---
In reply to: [133234248](https://github.com/Microsoft/clarity-js/pull/27#discussion_r133234248) [](ancestors = 133234248)

src/core.ts
ssivov: Refactored this for better practice and readability

---
In reply to: [133232916](https://github.com/Microsoft/clarity-js/pull/27#discussion_r133232916) [](ancestors = 133232916)

PR #28, opened 2017-08-31
Adding WebWorkers for compression

ssivov:


src/core.ts
```diff
@@ -255,10 +263,15 @@ function init() {
...
   droppedPayloads = {};
-  nextPayloadLength = 0;
   sentBytesCount = 0;
+  compressionWorker = createCompressionWorker(envelope(), onWorkerMessage);
+
+  // If critical API is missing, don't activate Clarity
+  if (!checkFeatures()) {
+    teardown();
+    return false;
+  }
```
ender336: We should probably do this first in the init, no need to create guids/workers just to tear them down right after

src/core.ts
```diff
@@ -84,21 +84,15 @@ export function addEvent(event: IEventData, scheduleUpload: boolean = true) {
...
-  // Infinite loop comes from sending instrumentation about failing to deliver previous delivery failure instrumentation.
-  let payloadIsSingleXhrErrorEvent = event.state && event.state.type === Instrumentation.XhrError && nextPayload.length === 1;
-  if (scheduleUpload && !payloadIsSingleXhrErrorEvent) {
+  let addEventMessage: IAddEventMessage = {
+    type: WorkerMessageType.AddEvent,
+    event: evt,
+    time: getTimestamp()
+  };
+  compressionWorker.postMessage(JSON.stringify(addEventMessage));
+  if (scheduleUpload) {
```
ender336: >scheduleUpload [](start = 6, length = 14)

what do you think about passing scheduleUpload to the worker and allowing the worker to handle setting these timeouts instead of managing it at this layer? In case setting and clearing timeouts ever starts to take a lot of time again we can make sure it is on the background thread and it seems like decisions of when to call forceUpload can stay within the worker isntead of us setting a timer to then send a message to the worker to forceUpload

src/core.ts
```diff
@@ -124,6 +126,24 @@ export function instrument(eventState: IInstrumentationEventState) {
...
+export function onWorkerMessage(evt: MessageEvent) {
+  let message = JSON.parse(evt.data) as IWorkerMessage;
+  switch (message.type) {
+    case WorkerMessageType.Upload:
+      let uploadMsg = message as IUploadMessage;
+      let onSuccess = (status: number) => { mapProperties(droppedPayloads, uploadDroppedPayloadsMappingFunction, true); };
+      let onFailure = (status: number) => { onFirstSendDeliveryFailure(status, uploadMsg.rawData, uploadMsg.compressedData); };
+      let compressedKb = Math.ceil(uploadMsg.compressedData.length / 1024.0);
+      let rawKb = Math.ceil(uploadMsg.rawData.length / 1024.0);
+      let envelope = JSON.parse(uploadMsg.rawData).envelope as IEnvelope;
```
ender336: nit: we are doing these calculations even if we aren't going to use them - can we inline them in the debug call to make sure we don't do any unnecessary work?

src/core.ts
```diff
@@ -124,6 +126,24 @@ export function instrument(eventState: IInstrumentationEventState) {
...
+      upload(uploadMsg.compressedData, onSuccess, onFailure);
+      debug(`** Clarity #${envelope.sequenceNumber}: Uploading ${compressedKb}KB (raw: ${rawKb}KB). **`);
+      break;
+    default:
```
ender336: >default [](start = 4, length = 7)

I think we should add some instrumentation to alert us if the worker is every giving us payloads that we aren't handling

src/compressionworker.ts
```diff
@@ -0,0 +1,97 @@
+import compress from "./compress";
+import { config } from "./config";
+
+export function createCompressionWorker(
+  sampleEnvelope: IEnvelope,
```
ender336: >sampleEnvelope [](start = 2, length = 14)

nit: sample is a confusing descriptor for me here. Is initialEnvelope a more accurate name perhaps?

ender336:
> let sequence: number;

it looks like we are handling sequence in the worker now and this only gets used to set the sampleEnvelope's sequence (which I don't think really gets used). Can we remove this and references to it?

---
Refers to: src/core.ts:18 in dd33993. [](commit_id = dd33993af22875fd32c9bbd9411fbbdbd9fdbcfd, deletion_comment = False)

src/core.ts
ssivov: In case check features fails, we are logging 'missing features' instrumentation and 'teardown' events. We need variables above to do that.

---
In reply to: [136670454](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136670454) [](ancestors = 136670454)

ssivov:
> let sequence: number;

Yeah, i think this can be refactored in some better way. I'll place in a change. Thanks !

---
In reply to: [326691970](https://github.com/Microsoft/clarity-js/pull/28#issuecomment-326691970) [](ancestors = 326691970)

---
Refers to: src/core.ts:18 in dd33993. [](commit_id = dd33993af22875fd32c9bbd9411fbbdbd9fdbcfd, deletion_comment = False)

test/compression.ts
```diff
@@ -1,4 +1,5 @@
 import compress from "../src/compress";
+import { cleanupFixture, setupFixture } from "./testsetup";
```
ender336: do we need this?

ssivov: Not in this suite: i'll remove this

---
In reply to: [136672535](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136672535) [](ancestors = 136672535)

src/core.ts
ssivov: How could this ever happen ?

---
In reply to: [136671497](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136671497) [](ancestors = 136671497)

src/core.ts
ssivov: Good catch

---
In reply to: [136671385](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136671385) [](ancestors = 136671385)

src/compressionworker.ts
ssivov: It's really the same object, that we re-use by updating sequence number for each next message - i think even just 'envelope' would be an ok name

---
In reply to: [136672217](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136672217) [](ancestors = 136672217)

src/core.ts
ender336: It shouldn't, but that's why we'd want to know if it did imo

---
In reply to: [136672665](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136672665) [](ancestors = 136672665,136671497)

test/errors.ts
```diff
@@ -21,8 +25,7 @@ describe("Error Tests", () => {
         syntheticEvent["lineno"] = lineno;
         syntheticEvent["colno"] = colno;
         errors.logError(syntheticEvent);
-        triggerSend();
-        let events = getAllSentEvents();
+        let events = stopObserving();
         let errorEvents = events.filter((event) => event.state.type === Instrumentation.JsError);
```
ender336: >let errorEvents = events.filter((event) => event.state.type === Instrumentation.JsError); [](start = 8, length = 89)

can the overserveEvents call now pass Instrumentation.JsError so we don't have to filter here?

ender336:
>         0,

nit and obviously unrelated to your change, but it would be nice to have a comment here for what the 0 param is given to

---
Refers to: test/pointer.ts:116 in dd33993. [](commit_id = dd33993af22875fd32c9bbd9411fbbdbd9fdbcfd, deletion_comment = False)

src/core.ts
ssivov: There is no other source to invoke this message handler, except from the code written in our own compression worker. I'd argue that placing a check here would be quite redundant

---
In reply to: [136673207](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136673207) [](ancestors = 136673207,136672665,136671497)

src/core.ts
ssivov: I thought about it at some point. I think my main hunch against it was design practice. I didn't think it'd be good for a worker to do something independently and be "smart" about it. I wanted to maintain the control of data flow within core and make worker act strictly based on the command messages.

---
In reply to: [136670940](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136670940) [](ancestors = 136670940)

test/errors.ts
ssivov: Yeah, i'll place the condition

---
In reply to: [136674238](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136674238) [](ancestors = 136674238)

ssivov:
>         0,

I'm working on PR related to pointers right now - i'll place the comment there :)

---
In reply to: [326695052](https://github.com/Microsoft/clarity-js/pull/28#issuecomment-326695052) [](ancestors = 326695052)

---
Refers to: test/pointer.ts:116 in dd33993. [](commit_id = dd33993af22875fd32c9bbd9411fbbdbd9fdbcfd, deletion_comment = False)

test/errors.ts
ssivov: Actually, you don't need any condition, since all of the new events should be error events only.

---
In reply to: [136677809](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136677809) [](ancestors = 136677809,136674238)

ssivov:
> let sequence: number;

After resolving Sarvesh's feedback, this variable becomes useful again, so will keep it in place

---
In reply to: [326692220](https://github.com/Microsoft/clarity-js/pull/28#issuecomment-326692220) [](ancestors = 326692220,326691970)

---
Refers to: src/core.ts:18 in dd33993. [](commit_id = dd33993af22875fd32c9bbd9411fbbdbd9fdbcfd, deletion_comment = False)

test/compressionworker.ts
```diff
@@ -0,0 +1,222 @@
...
+    secondMockEvent.state = { data: secondMockData };
+    let addEventMessage = createAddEventMessage(firstMockEvent);
+    let forceUploadMessage = createForceUploadMessage();
+
+    messageHandlers.push(messageHandler);
+    worker.postMessage(JSON.stringify(addEventMessage));
+    addEventMessage.event = secondMockEvent;
+    worker.postMessage(JSON.stringify(addEventMessage));
+    worker.postMessage(JSON.stringify(forceUploadMessage));
+    scheduleTestFailureTimeout();
```
toby-walker: Does this set up a timeout that causes the test to fail? Since this is a common pattern in these tests can you create a higher level abstraction that makes writing these easier? Lots of tedious boilerplate in these tests. Example:

Tests can be defined as a sequence of input events and output events and a worker. Write a generic test handler for that case.

test/compressionworker.ts
```diff
@@ -0,0 +1,222 @@
...
+      assert.equal(message.type, WorkerMessageType.Upload);
+      let uploadMessage = message as IUploadMessage;
+      let payload = JSON.parse(uploadMessage.rawData);
+      assert.equal(payload.events.length, 1);
+      assert.equal(payload.events[0].type, MockEventName);
+      done();
+    }
+  });
+
+  it("validates that payloads consisting of a single XHR error event are not uploaded", (done: DoneFn) => {
```
toby-walker: Should these never be uploaded? Doesn't it depend on state?

test/compressionworker.ts
```diff
@@ -0,0 +1,222 @@
...
+    let worker = createTestWorker();
+    let mockXhrErrorEvent = getMockEvent(InstrumentationEventName);
+    let mockXhrErrorEventState = {
+      type: Instrumentation.XhrError
+    } as IXhrErrorEventState;
+    mockXhrErrorEvent.state = mockXhrErrorEventState;
+    let addEventMessage: IAddEventMessage = {
+      type: WorkerMessageType.AddEvent,
+      time: new Date().getTime(),
```
toby-walker: Minor; Why not fix this value? One less changeable thing in the test.

test/compressionworker.ts
```diff
@@ -0,0 +1,222 @@
...
+    let forceUploadMessage = createForceUploadMessage();
+
+    messageHandlers.push(messageHandler);
+    worker.postMessage(JSON.stringify(addEventMessage));
+    worker.postMessage(JSON.stringify(forceUploadMessage));
+    scheduleTestSuccessTimeout(done);
+
+    function messageHandler(message: IWorkerMessage) {
+      // This handler should NOT be invoked, so intentional assertion failure in this path
+      assert.equal(true, false);
```
toby-walker: Isn't there a simple "fail" for this case?

test/compressionworker.ts
```diff
@@ -0,0 +1,222 @@
...
+    if (done) {
+      done();
+    }
+  }
+
+  function scheduleTestSuccessTimeout(done: DoneFn) {
+    jasmine.clock().uninstall();
+    testFailureTimeout = setTimeout(() => {
+      endTestByTimeout(true, done);
+    }, 1000);
```
toby-walker: Seems brittle.

src/core.ts
ender336: These sorts of comments are meant to make the codebase more robust against future changes. Even when you can prove that an unexpected state can't be reached in the current codebase - you can't guarantee that future changes won't unintentionally break your assumptions, especially in an open source project

---
In reply to: [136676138](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136676138) [](ancestors = 136676138,136673207,136672665,136671497)

test/compressionworker.ts
ssivov: Ok, I agree

---
In reply to: [136876641](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136876641) [](ancestors = 136876641)

ssivov: Changed it in all other mock events as well

---
In reply to: [137077163](https://github.com/Microsoft/clarity-js/pull/28#discussion_r137077163) [](ancestors = 137077163,136876641)

test/compressionworker.ts
ssivov: Yeah, turns out async 'done' object has fail method. Thanks !

---
In reply to: [136876665](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136876665) [](ancestors = 136876665)

test/compressionworker.ts
ssivov: This is an edge case. During normal Clarity runtime we don't sent a single XHR event because there is an infinite loop possibility:

SendEvent -> FailToDeliver -> Generate XHR error event -> Send XHR error event -> FailToDeliver -> Generate XHR error event -> Send XHR error event...etc.

On teardown, however, single XHR error event will be uploaded synchronously from core. I can add an extra test for that as well.

---
In reply to: [136876622](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136876622) [](ancestors = 136876622)

test/compressionworker.ts
ssivov: Why? 1 second should be safe upper limit for communication between treads in a test environment. I moved this value to be a constant at the top of the file.

---
In reply to: [136876709](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136876709) [](ancestors = 136876709)

src/core.ts
```diff
@@ -192,6 +184,14 @@ function upload(payload: string, onSuccess?: UploadCallback, onFailure?: UploadC
     defaultUpload(payload, onSuccess, onFailure);
   }
   sentBytesCount += payload.length;
+  if (state === State.Activated && sentBytesCount > config.totalLimit) {
```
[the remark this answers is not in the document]
ssivov: Because upload is invoked from teardown() method to upload remaining events. Is we don't check the state, then this would call circular call sequence.

---
In reply to: [136483683](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136483683) [](ancestors = 136483683)

src/core.ts
```diff
@@ -84,21 +84,15 @@ export function addEvent(event: IEventData, scheduleUpload: boolean = true) {
...
-  // This helps us avoid the infinite loop in the case when all requests fail (e.g. dropped internet connection)
-  // Infinite loop comes from sending instrumentation about failing to deliver previous delivery failure instrumentation.
-  let payloadIsSingleXhrErrorEvent = event.state && event.state.type === Instrumentation.XhrError && nextPayload.length === 1;
-  if (scheduleUpload && !payloadIsSingleXhrErrorEvent) {
+  let addEventMessage: IAddEventMessage = {
+    type: WorkerMessageType.AddEvent,
+    event: evt,
+    time: getTimestamp()
+  };
+  compressionWorker.postMessage(JSON.stringify(addEventMessage));
```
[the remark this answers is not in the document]
ssivov: Seems like no need to stringify for our types of objects - removed stringify calls.

---
In reply to: [136483494](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136483494) [](ancestors = 136483494)

src/compressionworker.ts
```diff
@@ -0,0 +1,97 @@
...
+  onMessage?: (e: MessageEvent) => void,
+  onError?: (e: ErrorEvent) => void
+): Worker {
+  let worker = null;
+  if (Worker) {
```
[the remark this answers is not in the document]
ssivov: Worker will be null and Clarity won't get activated because checkApi will return false

---
In reply to: [136482257](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136482257) [](ancestors = 136482257)

src/compressionworker.ts
```diff
@@ -0,0 +1,97 @@
...
+    }
+  }
+
+  function upload(compressed: string, uncompressed: string) {
+    let message: IUploadMessage = {
+      type: WorkerMessageType.Upload,
+      compressedData: compressed,
+      rawData: uncompressed
+    };
+    workerGlobalScope.postMessage(JSON.stringify(message));
```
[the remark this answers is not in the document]
ssivov: Posting message to the main thread

---
In reply to: [136483055](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136483055) [](ancestors = 136483055)

src/core.ts
```diff
@@ -255,10 +263,15 @@ function init() {
   startTime = getUnixTimestamp();
   activePlugins = [];
   bindings = {};
-  nextPayload = [];
   droppedPayloads = {};
-  nextPayloadLength = 0;
   sentBytesCount = 0;
+  compressionWorker = createCompressionWorker(envelope(), onWorkerMessage);
```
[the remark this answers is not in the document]
ssivov: There are a couple variables that need to be initialized to properly instrument teardown() call. I've split variable initializations in two parts: 

1. Variables required for teardown (left them up top)
2. Remaining variables (moved them to the bottom after all other checks)

---
In reply to: [136483845](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136483845) [](ancestors = 136483845)

src/compressionworker.ts
```diff
@@ -0,0 +1,97 @@
...
+      case WorkerMessageType.Terminate:
+        let terminatedMsg = message as ITimestampedWorkerMessage;
+        uploadNextPayload(terminatedMsg.time);
+        self.close();
+      default:
+        break;
+    }
+  };
+
+  function addEvent(event: IEvent, time: number) {
```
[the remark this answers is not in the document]
ssivov: That's true, but we need these functions to be easily stringified and bundled all together to run inside the worker scope. Pulling them in a separate file make passing this code inside the worker scope much harder.

---
In reply to: [136482705](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136482705) [](ancestors = 136482705)

src/compressionworker.ts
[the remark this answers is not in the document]
ssivov: XHR stuff remains in core's upload function. It stays in core for 2 reasons:
1. We need to be able to invoke custom upload handler, if such is specified in config
2. Monitoring dropped payloads and re-delivery logic is also defined in core. Later on I can pull that logic in a separate file/layer for cleaner code.

---
In reply to: [136483027](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136483027) [](ancestors = 136483027)

src/compressionworker.ts
```diff
@@ -0,0 +1,97 @@
...
+      envelope.time = time;
+      let raw = `{"envelope":${JSON.stringify(envelope)},"events":[${nextPayload.join()}]}`;
+      let compressed = compress(raw);
+      nextPayload = [];
+      nextPayloadLength = 0;
+      upload(compressed, raw);
+    }
+  }
+
+  function upload(compressed: string, uncompressed: string) {
```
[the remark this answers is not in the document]
ssivov: Worker just reports some data ready for upload to core. Upload logic and limit monitoring is done in core.

---
In reply to: [136482784](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136482784) [](ancestors = 136482784)

src/compressionworker.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  function addEvent(event: IEvent, time: number) {
+    let eventStr = JSON.stringify(event);
+    if (nextPayloadLength > 0 && nextPayloadLength + eventStr.length > config.batchLimit) {
+      uploadNextPayload(time);
+    }
+    nextPayload.push(eventStr);
+    nextPayloadLength += eventStr.length;
+    nextPayloadIsSingleXhrErrorEvent = (nextPayload.length === 1 && event.state && event.state.type === Instrumentation.XhrError);
+    if (nextPayloadLength >= config.batchLimit) {
```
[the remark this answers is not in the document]
ssivov: It would, but if we are already at the limit at this point, why wait until next call ? Added some comments too

---
In reply to: [136482622](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136482622) [](ancestors = 136482622)

src/compressionworker.ts
```diff
@@ -0,0 +1,97 @@
...
+      compressedData: compressed,
+      rawData: uncompressed
+    };
+    workerGlobalScope.postMessage(JSON.stringify(message));
+  }
+}
+
+function createWorkerUrl(sampleEnvelope: IEnvelope): string {
+  let workerContextStr = workerContext.toString();
+  let workerStr = workerContextStr.substring(workerContextStr.indexOf("{") + 1, workerContextStr.lastIndexOf("}"));
```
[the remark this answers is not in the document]
ssivov: Done

---
In reply to: [136483121](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136483121) [](ancestors = 136483121)

clarity.d.ts
```diff
@@ -293,7 +324,13 @@ declare const enum Instrumentation {
   Teardown,
   ClarityAssertFailed,
   ClarityDuplicated,
-  ShadowDomInconsistent
+  ShadowDomInconsistent,
+  Timestamp
+}
+
+declare const enum TimestampSource {
```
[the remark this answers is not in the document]
ssivov: I think this is outdated

---
In reply to: [136482090](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136482090) [](ancestors = 136482090)

src/compressionworker.ts
```diff
@@ -0,0 +1,97 @@
+import compress from "./compress";
+import { config } from "./config";
+
+export function createCompressionWorker(
```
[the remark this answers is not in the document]
ssivov: Do you mean 'return type'? If so, it is specified - 'Worker'

---
In reply to: [136482189](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136482189) [](ancestors = 136482189)

clarity.d.ts
```diff
@@ -355,6 +392,10 @@ interface IShadowDomInconsistentEventState extends IInstrumentationEventState {
   firstEvent?: IShadowDomInconsistentEventState;
 }
 
+interface ITimestampEventState extends IInstrumentationEventState {
```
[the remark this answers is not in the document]
ssivov: Again, this is probably outdated

---
In reply to: [136482138](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136482138) [](ancestors = 136482138)

src/compressionworker.ts
```diff
@@ -0,0 +1,97 @@
...
+    let message: IUploadMessage = {
+      type: WorkerMessageType.Upload,
+      compressedData: compressed,
+      rawData: uncompressed
+    };
+    workerGlobalScope.postMessage(JSON.stringify(message));
+  }
+}
+
+function createWorkerUrl(sampleEnvelope: IEnvelope): string {
```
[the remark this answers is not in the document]
ssivov: Essentially, yes. Worker requires a URL in its constructor, pointing to the code, which it will run. This creates a blob, containing a 'bundle' with ALL required worker routines, which would be executed in a completely isolated worker's scope and returns a URL to this Blob. 

---
In reply to: [136483183](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136483183) [](ancestors = 136483183)

src/compressionworker.ts
```diff
@@ -0,0 +1,97 @@
...
+  let workerContextStr = workerContext.toString();
+  let workerStr = workerContextStr.substring(workerContextStr.indexOf("{") + 1, workerContextStr.lastIndexOf("}"));
+  let compressStr = `self.compress=${compress.toString()};`;
```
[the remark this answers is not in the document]
ssivov: Could you please expand a little ? Not clear to me what you mean here

---
In reply to: [136483225](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136483225) [](ancestors = 136483225)

src/compressionworker.ts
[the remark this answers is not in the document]
toby-walker: A future refactor we might want to create a worker for sending the (already) compressed data.

---
In reply to: [137139738](https://github.com/Microsoft/clarity-js/pull/28#discussion_r137139738) [](ancestors = 137139738,136483027)

src/compressionworker.ts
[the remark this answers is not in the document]
toby-walker: Something more like this:

  let workerContextStr = workerContext.toString();
  let workerStr = workerContextStr.substring(workerContextStr.indexOf("{") + 1, workerContextStr.lastIndexOf("}"));
  let envelopeStr = `self.envelope=${JSON.stringify(sampleEnvelope)};`;
  let code = `
    self.compress=${compress.toString()}
    self.config=${JSON.stringify(config)};
    self.envelope=${JSON.stringify(envelope)}
    ${workerStr}
  `;



---
In reply to: [137143606](https://github.com/Microsoft/clarity-js/pull/28#discussion_r137143606) [](ancestors = 137143606,136483225)

src/compressionworker.ts
```diff
@@ -0,0 +1,105 @@
...
+    }
+  };
+
+  function addEvent(event: IEvent, time: number): void {
+    let eventStr = JSON.stringify(event);
+
+    // If appending new event to next payload would exceed batch limit, then flush next payload first
+    if (nextPayloadBytes > 0 && nextPayloadBytes + eventStr.length > config.batchLimit) {
+      uploadNextPayload(time);
+    }
```
toby-walker: Will we ever stop uploading? It's not clear from this logic it will do so. Better would be to add a state to indicate "stop uploading" and set that once and use here to bail out quickly.

src/compressionworker.ts
```diff
@@ -0,0 +1,105 @@
...
+  let workerGlobalScope = self as any;
+  let compress = workerGlobalScope.compress;
+  let config = workerGlobalScope.config;
+  let envelope: IEnvelope = workerGlobalScope.envelope;
+  let nextPayloadEvents: IEvent[] = [];
+  let nextPayloadBytes = 0;
+  let sequence = 0;
+
+  // Edge case: Flag to skip uploading payloads consisting of a single XhrError instrumentation event
+  // This helps us avoid the infinite loop in the case when all requests fail (e.g. dropped internet connection)
```
toby-walker: >infinite [](start = 29, length = 8)

Could we instead change the XHR callback to hit a maximum number of retires? This is separate from your changes.

src/core.ts
```diff
@@ -60,11 +62,14 @@ export function teardown() {
   }
 
   delete document[ClarityAttribute];
+  if (compressionWorker) {
+    compressionWorker.terminate();
```
toby-walker: >terminate [](start = 22, length = 9)

Will this lose work in progress in the worker if we haven't gotten all messages back yet?

src/core.ts
```diff
@@ -60,11 +62,14 @@ export function teardown() {
...
+  }
   state = State.Unloaded;
 
-  // Upload residual events
+  // Instrument teardown and upload residual events
   instrument({ type: Instrumentation.Teardown });
-  uploadNextPayload();
+  uploadPendingEvents();
```
toby-walker: >uploadPendingEvents [](start = 2, length = 19)

How does compression happen for these if you terminate the worker above?

src/core.ts
```diff
@@ -20,9 +21,10 @@ let startTime: number;
 let activePlugins: IPlugin[];
 let bindings: IBindingContainer;
 let droppedPayloads: { [key: number]: IDroppedPayloadInfo };
+let pendingEvents: IEvent[] = [];
```
toby-walker: >pendingEvents [](start = 4, length = 13)

Can you add a comment about what this is doing?

src/core.ts
```diff
@@ -124,6 +132,30 @@ export function instrument(eventState: IInstrumentationEventState) {
...
+export function onWorkerMessage(evt: MessageEvent) {
+  if (state !== State.Unloaded) {
+    let message = evt.data;
+    switch (message.type) {
+      case WorkerMessageType.Upload:
+        let uploadMsg = message as IUploadMessage;
+        let onSuccess = (status: number) => { mapProperties(droppedPayloads, uploadDroppedPayloadsMappingFunction, true); };
+        let onFailure = (status: number) => { onFirstSendDeliveryFailure(status, uploadMsg.rawData, uploadMsg.compressedData); };
+        upload(uploadMsg.compressedData, onSuccess, onFailure);
+        pendingEvents.splice(0, uploadMsg.eventCount);
```
toby-walker: >     pendingEvents.splice(0, uploadMsg.eventCount); [](start = 3, length = 51)

Are you guaranteed that the state of this will always match the messages back from the worker? Seems a bit brittle. I would at least add some checking here that they align.

toby-walker: >splice [](start = 22, length = 6)

Suggest is to have something here that lets you see if it matches what is in the messages posted back from the worker and check the match.

src/compressionworker.ts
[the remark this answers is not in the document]
toby-walker: Could you add a comment that basically include your response above?

---
In reply to: [137143542](https://github.com/Microsoft/clarity-js/pull/28#discussion_r137143542) [](ancestors = 137143542,136483183)

src/core.ts
ssivov: We compress this in the main thread for teardown (for example on 'unload') on purpose

---
In reply to: [137239714](https://github.com/Microsoft/clarity-js/pull/28#discussion_r137239714) [](ancestors = 137239714)

src/core.ts
ssivov: Yes, 'terminate' kills the worker immediately. There might be messages that were already posted by the worker, but were not received by main thread yet, but we'll discard those based on the unloaded state.

---
In reply to: [137239598](https://github.com/Microsoft/clarity-js/pull/28#discussion_r137239598) [](ancestors = 137239598)

src/compressionworker.ts
ssivov: This infinite loop doesn't come from re-uploading dropped payloads. It comes from generating new event for a failed send and sending that as a new payload, so it's a different issue.

---
In reply to: [137239428](https://github.com/Microsoft/clarity-js/pull/28#discussion_r137239428) [](ancestors = 137239428)

src/compressionworker.ts
ssivov: Why would we not stop uploading? We are uploading at most 1 batch here. I think i should rename these functions to 'sendToMain' or something like that to avoid confusion, since these calls aren't doing the actual uploading.

---
In reply to: [137238671](https://github.com/Microsoft/clarity-js/pull/28#discussion_r137238671) [](ancestors = 137238671)

src/core.ts
ssivov: Yes, as long as there is only one worker. The order of messages is guaranteed, so worker will be compressing and sending back events in the exact same order as pushed them on the queue.

---
In reply to: [137240383](https://github.com/Microsoft/clarity-js/pull/28#discussion_r137240383) [](ancestors = 137240383)

src/compressionworker.ts
ssivov: Renamed these functions and updated comments

---
In reply to: [137354749](https://github.com/Microsoft/clarity-js/pull/28#discussion_r137354749) [](ancestors = 137354749,137238671)

src/core.ts
ssivov: Done

---
In reply to: [137239828](https://github.com/Microsoft/clarity-js/pull/28#discussion_r137239828) [](ancestors = 137239828)

src/core.ts
ssivov: I could store pendingEvents as an object Map<eventId, eventObject>, and then parse batch returned by the worker to inspect all event IDs that it contains and then clear them from the pendingEvents storage 1 by 1.

This would be less brittle, but obviously more complicated.

---
In reply to: [137240630](https://github.com/Microsoft/clarity-js/pull/28#discussion_r137240630) [](ancestors = 137240630)

src/compressionworker.ts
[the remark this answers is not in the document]
ssivov: Added it above the 'createWorkerUrl' function

---
In reply to: [137241154](https://github.com/Microsoft/clarity-js/pull/28#discussion_r137241154) [](ancestors = 137241154,137143542,136483183)

src/compressionworker.ts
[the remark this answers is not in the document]
ssivov: I didn't find a way to pass a custom upload handler to the worker (especially if that handler invokes some separately defined subroutine)

---
In reply to: [137234561](https://github.com/Microsoft/clarity-js/pull/28#discussion_r137234561) [](ancestors = 137234561,137139738,136483027)

src/compressionworker.ts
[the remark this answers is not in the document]
ssivov: Ok, done

---
In reply to: [137238004](https://github.com/Microsoft/clarity-js/pull/28#discussion_r137238004) [](ancestors = 137238004,137143606,136483225)

clarity.d.ts
```diff
@@ -105,6 +105,37 @@ type UploadCallback = (status: number) => void;
...
+  /* Main thread to Worker messages */
+  AddEvent,
+  ForceUpload,
+  Terminate,
+
+  /* Worker to main thread messages */
+  Upload
+}
+
+interface IWorkerMessage {
```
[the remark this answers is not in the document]
ssivov: They require literal discriminators that are not suitable for .d.ts file

---
In reply to: [136482038](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136482038) [](ancestors = 136482038)

test/compressionworker.ts
ssivov: Ok, so I played around with this idea, but it doesn't look like it has much gain. Tests have custom scenarios and can't ALL be placed in the same exact abstraction. I don't think there are many enough tests in this suite to justify the extra work.

---
In reply to: [136876565](https://github.com/Microsoft/clarity-js/pull/28#discussion_r136876565) [](ancestors = 136876565)

src/compressionworker.ts
toby-walker: Understood it's a different issue. Just wondering if we should fix the "never ending retry" problem (not the infinite loop) if easy as part of this PR. 

---
In reply to: [137353755](https://github.com/Microsoft/clarity-js/pull/28#discussion_r137353755) [](ancestors = 137353755,137239428)

src/core.ts
toby-walker: Can you clarify what you mean "discard those based on unloaded state". Maybe put the answer in a comment here. 

---
In reply to: [137352823](https://github.com/Microsoft/clarity-js/pull/28#discussion_r137352823) [](ancestors = 137352823,137239598)

src/core.ts
toby-walker: Can you add that as comment? 

---
In reply to: [137355388](https://github.com/Microsoft/clarity-js/pull/28#discussion_r137355388) [](ancestors = 137355388,137240383)

src/compressionworker.ts
ssivov: I think that making a separate PR, which would split off the sending/status tracking/re-seding logic in a separate entity, would be a better fit for that change. (I was going to make another PR for that, regardless)

---
In reply to: [137421412](https://github.com/Microsoft/clarity-js/pull/28#discussion_r137421412) [](ancestors = 137421412,137353755,137239428)

src/core.ts
ssivov: 'onWorkerMessage' message handler function starts with 'if (state !== State.Unloaded)', so all events that are delivered after teardown will be ignored.

---
In reply to: [137421841](https://github.com/Microsoft/clarity-js/pull/28#discussion_r137421841) [](ancestors = 137421841,137352823,137239598)

src/core.ts
ssivov: Added

---
In reply to: [137421997](https://github.com/Microsoft/clarity-js/pull/28#discussion_r137421997) [](ancestors = 137421997,137355388,137240383)

src/core.ts
ssivov: Okay.

---
In reply to: [137422665](https://github.com/Microsoft/clarity-js/pull/28#discussion_r137422665) [](ancestors = 137422665,137421841,137352823,137239598)

PR #29, opened 2017-09-04
Improving box model support for clarity

sarveshnagpal:
Generalizing box model visualization. Two high-level changes:
1) Do not capture box model for elements with hidden visibility
2) Capture foreground color to assist with visualizing generic box models

### DIFF
--- a/clarity.d.ts
+++ b/clarity.d.ts
@@ -70,11 +70,15 @@ interface IEnvelope {
   sequenceNumber: number;
 }
 
-interface IEvent {
+interface IEventData {
+  type: string;
+  state: any;
+  time?: number;
+}
+
+interface IEvent extends IEventData {
   id: number; /* Event ID */
   time: number; /* Time relative to page start */
-  type: string; /* Type of the event */
-  state: any;
 }
 
 interface IDroppedPayloadInfo {
--- a/clarity.d.ts
+++ b/clarity.d.ts
@@ -21,6 +21,12 @@ interface IConfig {
   // If set to false, src of images won't be captured
   showImages?: boolean;
 
+  // If set to false, box model won't capture lines for text nodes
+  showLines?: boolean;
+
+  // If set to false, box model won't capture the foreground color of the element
+  fetchColor?: boolean;
+
   // Maximum number of milliseconds, after which Clarity should yield the thread
   // It is used to avoid freezing the page during large object serialization
   timeToYield?: number;
@@ -86,6 +92,12 @@ interface IDroppedPayloadInfo {
   xhrErrorState: IXhrErrorEventState;
 }
 
+interface IJsProfiler {
+  calls: number;
+  duration: number;
+  start: number;
+}
+
 interface IPlugin {
   activate(): void;
   teardown(): void;
@@ -122,10 +134,11 @@ interface IBoxModel {
   y: number; /* Y coordinate of the element */
   width: number; /* Width of the element */
   height: number; /* Height of the element */
-  color: string; /* Color for visual aid */
   scrollX?: number; /* Scroll left of the element */
   scrollY?: number; /* Scroll top of the element */
-  overflow?: string; /* Overflow property for visualization aid */
+  overflow?: string; /* Overflow property for visualization */
+  visibility?: string; /* Visibility property for visualization */
+  color?: string; /* Color to assist with visualizing box model */
 }
 
 declare const enum Source {
@@ -182,6 +195,7 @@ interface IElementLayoutState extends ILayoutState {
 
 interface ITextLayoutState extends ILayoutState {
   content: string;
+  lines?: IBoxModel[];
 }
 
 interface IIgnoreLayoutState extends ILayoutState {
@@ -295,7 +309,8 @@ declare const enum Instrumentation {
   Teardown,
   ClarityAssertFailed,
   ClarityDuplicated,
-  ShadowDomInconsistent
+  ShadowDomInconsistent,
+  JsProfile
 }
 
 interface IInstrumentationEventState {
@@ -308,6 +323,12 @@ interface IJsErrorEventState extends IInstrumentationEventState {
   stack: string;
   lineno: number;
   colno: number;
+}
+
+interface IJsProfileEventState extends IInstrumentationEventState {
+  label: string;
+  calls: number;
+  duration: number;
 }
 
 interface IMissingFeatureEventState extends IInstrumentationEventState {
--- a/clarity.d.ts
+++ b/clarity.d.ts
@@ -117,13 +117,15 @@ interface IShadowDomNode extends HTMLDivElement {
   ignore: boolean;  /* Flag to avoid sending data for that node */
 }
 
-interface ILayoutRectangle {
+interface IBoxModel {
   x: number; /* X coordinate of the element */
   y: number; /* Y coordinate of the element */
   width: number; /* Width of the element */
   height: number; /* Height of the element */
+  color: string; /* Color for visual aid */
   scrollX?: number; /* Scroll left of the element */
   scrollY?: number; /* Scroll top of the element */
+  overflow?: string; /* Overflow property for visualization aid */
 }
 
 declare const enum Source {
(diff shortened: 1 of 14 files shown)


src/plugins/layout.ts
```diff
@@ -166,12 +168,13 @@ export default class Layout implements IPlugin {
 
   private createEventState<T extends ILayoutEventInfo>(eventInfo: T): ILayoutState {
     let node = eventInfo.node;
-    let layoutState: ILayoutState = createLayoutState(node, this.shadowDom);
+    let styles = node.nodeType === Node.ELEMENT_NODE ? window.getComputedStyle(node as Element) : null;
```
ssivov: >let styles = node.nodeType === Node.ELEMENT_NODE ? window.getComputedStyle(node as Element) : null; [](start = 4, length = 99)

This is going to be expensive. Can we only do it for inserts and styles updates? Also, when stylesheet is inserted, would this mean you'd need to recompute styles for all nodes ?

clarity.d.ts
```diff
@@ -117,13 +117,15 @@ interface IShadowDomNode extends HTMLDivElement {
   ignore: boolean;  /* Flag to avoid sending data for that node */
 }
 
-interface ILayoutRectangle {
+interface IBoxModel {
   x: number; /* X coordinate of the element */
   y: number; /* Y coordinate of the element */
   width: number; /* Width of the element */
   height: number; /* Height of the element */
+  color: string; /* Color for visual aid */
```
ssivov: >color [](start = 2, length = 5)

Why do we need color specifically ?

src/plugins/layout.ts
sarveshnagpal: For correctness of box model, re-valuating dimensions is necessary and can't be behind a configuration. In other words box model as part of clarity should never be optional. 

Your point is valid, elements don't necessarily get a valid mutation event for CSS updates and therefore will need to be re-evaluated every time we get a mutation in the parent tree or addition of style sheet. What I can probably do to reduce the work here is to not capture box model for every element and instead limit collection to elements that we will need for the analysis.

clarity.d.ts
sarveshnagpal: For corresponding visualization changes in clarity-tools (you'll see a PR there soon). Although, this is optional and only for visualization aid. The reason I grabbed it was because we already had access to computed styles because we have to check "visibility" property and it was a cheap lookup.

toby-walker: What's the format of color?

clarity.d.ts
```diff
@@ -117,13 +117,15 @@ interface IShadowDomNode extends HTMLDivElement {
...
   scrollX?: number; /* Scroll left of the element */
   scrollY?: number; /* Scroll top of the element */
+  overflow?: string; /* Overflow property for visualization aid */
```
toby-walker: What does overflow mean though?

src/plugins/layout.ts
```diff
@@ -115,7 +115,9 @@ export default class Layout implements IPlugin {
     while (this.originalLayouts.length > 0 && getTimestamp(true) < yieldTime) {
       let originalLayout = this.originalLayouts.shift();
       let originalLayoutState = originalLayout.layout;
-      let currentLayoutState = createLayoutState(originalLayout.node, this.shadowDom);
+      let originalNode = originalLayout.node;
+      let currentStyles = originalNode.nodeType === Node.ELEMENT_NODE ? window.getComputedStyle(originalNode as Element) : null;
+      let currentLayoutState = createLayoutState(originalNode, currentStyles, this.shadowDom);
```
toby-walker: Isn't this expensive? Do we need it everywhere? Seems a bit high cost for the value.

src/plugins/layout.ts
toby-walker: Do you need to evaluate all computed styles? Can you get only those you need by some alternate, faster means? Can we also profile the impact of this change before profile it? If we don't have it, let's add perf instrumentation to Clarity to make sure we can measure this kind of impact (for this change or any other).

src/plugins/layout.ts
```diff
@@ -198,15 +201,14 @@ export default class Layout implements IPlugin {
...
 
     // We only wish to watch elements once and then wait on the events to push changes
     if (this.watchList[layoutState.index]) {
       return;
     }
 
     // Check if scroll is possible, and if so, bind to scroll event
-    let styles = window.getComputedStyle(element);
     let scrollPossible = (layoutState.layout !== null
                           && (styles["overflow-x"] === "auto"
```
toby-walker: Possible null dereference here  if styles == null. Didn't check the entire code to see if that was the case.

src/plugins/layout/stateprovider.ts
```diff
@@ -89,12 +89,20 @@ export function createElementLayoutState(element: Element): IElementLayoutState
 
   elementState.layout = null;
   if (rect) {
-    elementState.layout = {
-      x: Math.round(rect.left),
-      y: Math.round(rect.top),
-      width: Math.round(rect.width),
-      height: Math.round(rect.height)
-    };
+    if (styles["visibility"] !== "hidden") {
```
toby-walker: Why do you not capture the visibility style value explicitly rather and send that in Clarity?

src/plugins/layout/stateprovider.ts
```diff
@@ -89,12 +89,20 @@ export function createElementLayoutState(element: Element): IElementLayoutState
...
+      elementState.layout = {
+        x: Math.round(rect.left),
+        y: Math.round(rect.top),
+        width: Math.round(rect.width),
+        height: Math.round(rect.height),
+        color: styles["color"]
+       };
```
toby-walker: Could you not jsut do:

node.style.color and similarly for other cases to compute only the CSS properties you need and not everything?

src/plugins/layout/stateprovider.ts
```diff
@@ -89,12 +89,20 @@ export function createElementLayoutState(element: Element): IElementLayoutState
...
   if (rect) {
-    elementState.layout = {
-      x: Math.round(rect.left),
-      y: Math.round(rect.top),
-      width: Math.round(rect.width),
-      height: Math.round(rect.height)
-    };
+    if (styles["visibility"] !== "hidden") {
+      elementState.layout = {
+        x: Math.round(rect.left),
```
toby-walker: Why do we round? How often do we miss thing with fractional pixels (on phones, etc.)?

src/plugins/layout.ts
sarveshnagpal: Unfortunately there's no faster way other than accessing all computed styles. E.g. element.style.visibility is much faster however it will only return a valid value if there's an inline style. For styles via CSS it doesn't work. 

Yes, added code for profiling. That was a good idea. On my machine we are averaging 50 operations per millisecond and I've made changes to only grab layout data for leaf nodes.

clarity.d.ts
sarveshnagpal: It's rgba. Based on feedback, I've put it behind a config so we don't have to worry about it in most cases but it does help at the time of a demo.

clarity.d.ts
sarveshnagpal: It means that the parent element can hide any portions of the children that exceed its specified dimensions. This is a common technique used in CSS image spriting where an actual image contains a bunch of images and using CSS we control what's visible to the user. Without this box model will show up as incorrect.

src/plugins/layout.ts
sarveshnagpal: Not necessarily - the operation is cheap and is something we were already doing for majority of computations when calculating if the element supports scrolling or not. I've added code to measure performance.

src/plugins/layout.ts
sarveshnagpal: I've reverted these lines to keep things simple. And do we need to evaluate all styles? Unfortunately yes, no other straight forward way to compute only one property.  And already profiling in the new iteration.

src/plugins/layout/stateprovider.ts
sarveshnagpal: Good point, now sending it as part of Clarity.

src/plugins/layout/stateprovider.ts
sarveshnagpal: We round for pixel level precision to save on bytes that we transfer. We can discuss this offline if we would like to change the concept to expand from integers to floats but for now this is what we have been using.

src/plugins/layout/stateprovider.ts
sarveshnagpal: node.style.color only works if the color property is assigned via inline style. If a style sheet was used, it gives inaccurate results. To factor in cascading nature of style sheets, we will need computed styles.

PR #35, opened 2017-09-21
Bug fix and monitoring scroll for overflow hidden elements

sarveshnagpal:
This code change fixes the bug where we were only setting scroll position the first time element was inserted. If we received a follow-up mutation, then we lost the existing scroll position even though we continued to watch the element for scroll updates. If at this time scroll event happened, we would try to check if distance is large enough for us to send the event back and we would see previous scroll position as NaN. 

The other fix in this code change is to also watch scrolling for overflow:hidden elements.

### DIFF
--- a/src/plugins/layout/stateprovider.ts
+++ b/src/plugins/layout/stateprovider.ts
@@ -89,12 +89,25 @@ export function createElementLayoutState(element: Element): IElementLayoutState 
 
   elementState.layout = null;
   if (rect) {
+    let styles = window.getComputedStyle(element);
+
     elementState.layout = {
       x: Math.round(rect.left),
       y: Math.round(rect.top),
       width: Math.round(rect.width),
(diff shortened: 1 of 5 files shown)


src/plugins/layout/stateprovider.ts
```diff
@@ -89,12 +89,25 @@ export function createElementLayoutState(element: Element): IElementLayoutState
...
+    let styles = window.getComputedStyle(element);
+
     elementState.layout = {
       x: Math.round(rect.left),
       y: Math.round(rect.top),
       width: Math.round(rect.width),
       height: Math.round(rect.height)
     };
+
+    // Check if scroll is possible, and if so, bind to scroll event
```
ssivov: >// Check if scroll is possible, and if so, bind to scroll event [](start = 4, length = 63)

Outdated comment

PR #39, opened 2017-10-12
 Removing redundant data from envelope

ssivov:


clarity.d.ts
```diff
@@ -73,12 +73,16 @@ interface IPayload {
 }
 
 interface IEnvelope {
+  impressionId: string;
+  time?: number;
+  sequenceNumber?: number;
+}
+
+interface IClientInfo {
```
sarveshnagpal: Have you considered keeping all this info in envelope except not repeating redundant stuff in subsequent envelopes to save on bytes. This will ensure that if for some reason top URL changes (like # appended) we will still capture it but won't waste bytes if nothing changed?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clarity-js",
-  "version": "0.1.16",
+  "version": "0.2.0",
```
sarveshnagpal: That's a jump... ambitious pull request :)

src/core.ts
```diff
@@ -47,8 +48,12 @@ export function activate() {
     init();
   } catch (e) {
     onActivateErrorUnsafe(e);
+    return;
   }
 
+  // Send client information as the first event
+  sendClientInfo();
```
sarveshnagpal: Based on my earlier comment, let's see if we can avoid adding another top-level JSON object. Envelope is (or was) supposed to address this.

package.json
ssivov: I moved id up because there is a data format change

---
In reply to: [144364548](https://github.com/Microsoft/clarity-js/pull/39#discussion_r144364548) [](ancestors = 144364548)

clarity.d.ts
ssivov: I have, but i thought that it wasn't clear enough. If we did that, we would be relying on previous payloads constantly and would have to re-scan entire data from start to end to see what was going on. Everything, except in rare cases url, is constant in this object. If we think it's important to track url changes, we should probably fire a separtate 'Redirect' event.

---
In reply to: [144364338](https://github.com/Microsoft/clarity-js/pull/39#discussion_r144364338) [](ancestors = 144364338)

src/core.ts
ssivov: Please see my response to your previous comment

---
In reply to: [144364782](https://github.com/Microsoft/clarity-js/pull/39#discussion_r144364782) [](ancestors = 144364782)

clarity.d.ts
```diff
@@ -73,12 +73,16 @@ interface IPayload {
...
   clarityId: string;
   impressionId: string;
   url: string;
   version: string;
-  time?: number;
-  sequenceNumber?: number;
```
toby-walker: Why are these nullable? I realize that is before this change but curious why.

src/core.ts
```diff
@@ -58,9 +63,11 @@ export function activate() {
       state = State.Activated;
     } else {
       teardown();
+      return;
     }
   } catch (e) {
     onActivateError(e);
+    return;
```
toby-walker: >return [](start = 4, length = 6)

What does adding these fix? Seems unrelated to format change.

src/core.ts
```diff
@@ -358,6 +360,20 @@ function prepare() {
   return true;
 }
 
+function sendClientInfo() {
```
toby-walker: >sendClientInfo [](start = 9, length = 14)

PageData seems better to me for a name.

clarity.d.ts
ssivov: All other properties in the envelope are constant throughout the impression, so we were storing an "abstract" envelope object without time and sequenceNo, which was later used in various places by appending relevant time and sequenceNo to it. 

This behavior is optimized now to avoid this pattern.

---
In reply to: [144390698](https://github.com/Microsoft/clarity-js/pull/39#discussion_r144390698) [](ancestors = 144390698)

src/core.ts
ssivov: It is unrelated to the format change, but it was a small fix that I just corrected on the fly, while working on this PR :P

Essentially each of these catch blocks means that we've ran into a critical problem. These added 'return' statements make sure we terminate immediately after handling the problem

---
In reply to: [144390879](https://github.com/Microsoft/clarity-js/pull/39#discussion_r144390879) [](ancestors = 144390879)

toby-walker: I think this branch will *not* be merged for a while. Can you make these same fixes to the main (beta) version?

---
In reply to: [144425231](https://github.com/Microsoft/clarity-js/pull/39#discussion_r144425231) [](ancestors = 144425231,144390879)

package.json
toby-walker: I agree this is an appropriate bump. We will probably also go through multiple version on the "main" 0.1 series before this one goes out too. 

---
In reply to: [144371478](https://github.com/Microsoft/clarity-js/pull/39#discussion_r144371478) [](ancestors = 144371478,144364548)

clarity.d.ts
ssivov: Synced offline: We renamed this to 'metadata' and we'd be passing it as a part of the first payload.

First payload will have the following structure:
{
  envelope: IEnvelope;
  events: IEvent[];
  metadata: IImpressionMetadata;
}

Subsequent payloads will just contain envelope and events.

---
In reply to: [144372367](https://github.com/Microsoft/clarity-js/pull/39#discussion_r144372367) [](ancestors = 144372367,144364338)

src/core.ts
ssivov: Published a separate PR for that

---
In reply to: [144539040](https://github.com/Microsoft/clarity-js/pull/39#discussion_r144539040) [](ancestors = 144539040,144425231,144390879)

src/core.ts
ssivov: Synced offline: We renamed this to 'metadata' and we'd be passing it as a part of the first payload.

First payload will have the following structure:
{
  envelope: IEnvelope;
  events: IEvent[];
  metadata: IImpressionMetadata;
}

Subsequent payloads will just contain envelope and events.

---
In reply to: [144372712](https://github.com/Microsoft/clarity-js/pull/39#discussion_r144372712) [](ancestors = 144372712,144364782)

src/core.ts
ssivov: It also contains clientId - let's call it metadata

---
In reply to: [144391024](https://github.com/Microsoft/clarity-js/pull/39#discussion_r144391024) [](ancestors = 144391024)

PR #44, opened 2017-11-01
Data format changes to reduce payloads

ssivov:


converters/toarray.ts
```diff
@@ -0,0 +1,26 @@
+import DiscoverConverters from "./toarray/discover";
+import InstrumentationConverters from "./toarray/instrumentation";
+import LayoutConverters from "./toarray/layout";
+import PerformanceConverters from "./toarray/performance";
+import PointerConverters from "./toarray/pointer";
+import ViewportConverters from "./toarray/viewport";
+
+export default function(event: IEvent): IEventArray {
```
toby-walker: >function(event: IEvent): IEventArray [](start = 15, length = 36)

I would prefer to have an interface /serializer. Something like:

IEventSerializer<TIn,Tout> IEventSerializer
{
    TOut  serialize(TIn input)
   TIn deserailize(TOut output)
}

Seems cleaner and easier to understand. But keep the binding very high level.

converters/convert.ts
```diff
@@ -0,0 +1,5 @@
+import fromarray from "./fromarray";
+import { discoverToEvents } from "./fromarray/discover";
+import toarray from "./toarray";
+
+export default { fromarray, toarray, discoverToEvents };
```
toby-walker: >fromarray [](start = 17, length = 9)

Don't bke the details in this so much. Just think of "in and out" Then you have the converters going in and out. See comment about defining an interface to the conversion to/from.

converters/fromarray.ts
```diff
@@ -0,0 +1,24 @@
...
+  let origin  = eventArray[1];
+  let type    = eventArray[2];
+  let time    = eventArray[3];
+  let data    = converters[origin][type](eventArray[4]);
+  let event: IEvent = { origin, type, id, time, data };
+  return event;
+}
+
+let converters = [];
+converters[Origin.Discover] = DiscoverConverters;
```
toby-walker: >Discover [](start = 18, length = 8)

What does "Discover"mean?

converters/toarray.ts
```diff
@@ -0,0 +1,26 @@
...
+    event.origin,
+    event.type,
+    event.time,
+    converter(event.data)
+  ];
+  return data;
+}
+
+let converters = [];
+converters[Origin.Discover] = DiscoverConverters;
```
toby-walker: >DiscoverConverters [](start = 30, length = 18)

Names here are confusing as the direction of the conversion isn't clear.

converters/toarray/discover.ts
```diff
@@ -0,0 +1,9 @@
+let discoverConverters = [];
+discoverConverters[DiscoverEventType.Discover] = discoverToArray;
+
+export default discoverConverters;
+
+function discoverToArray(discover: IDiscover) {
```
toby-walker: >discoverToArray(discover: IDiscove [](start = 9, length = 34)

Add typing for result. Also can't these be defined as an interface

toby-walker:
> 

This looks like a dispatch table. How about making this a set of classes and using typing instead?

---
Refers to: converters/toarray.ts:27 in 6443c3d. [](commit_id = 6443c3dca3249ffed3a66b5f0175f2f0855835a4, deletion_comment = False)

converters/fromarray/instrumentation.ts
```diff
@@ -0,0 +1,90 @@
...
+instrumentationConverters[Instrumentation.JsError] = jsErrorFromArray;
+instrumentationConverters[Instrumentation.MissingFeature] = missingFeatureFromArray;
+instrumentationConverters[Instrumentation.XhrError] = xhrErrorFromArray;
+instrumentationConverters[Instrumentation.TotalByteLimitExceeded] = byteLimitExceededFromArray;
+instrumentationConverters[Instrumentation.Teardown] = noDataFromArray;
+instrumentationConverters[Instrumentation.PerformanceStateError] = noDataFromArray;
+instrumentationConverters[Instrumentation.ClarityAssertFailed] = assertFailedFromArray;
+instrumentationConverters[Instrumentation.ClarityDuplicated] = clarityDuplicatedFromArray;
+instrumentationConverters[Instrumentation.ClarityActivateError] = clarityActivateErrorFromArray;
+instrumentationConverters[Instrumentation.ShadowDomInconsistent] = inconsistentShadowDomFromArray;
```
toby-walker: This is a bunch of binding. Would prefer classes with subtypes instead.

converters/fromarray/instrumentation.ts
```diff
@@ -0,0 +1,90 @@
...
+
+export default instrumentationConverters;
+
+function jsErrorFromArray(jsErrorData: any[]): IJsErrorEventData {
+  let data: IJsErrorEventData = {
+    source  : jsErrorData[0],
+    message : jsErrorData[1],
+    stack   : jsErrorData[2],
+    lineno  : jsErrorData[3],
+    colno   : jsErrorData[4]
```
toby-walker: Could you not encode this mapping in a data structure and make the code generic? Or code-gen it if you are concerned about performance.

converters/fromarray/instrumentation.ts
```diff
@@ -0,0 +1,90 @@
...
+instrumentationConverters[Instrumentation.Teardown] = noDataFromArray;
+instrumentationConverters[Instrumentation.PerformanceStateError] = noDataFromArray;
+instrumentationConverters[Instrumentation.ClarityAssertFailed] = assertFailedFromArray;
+instrumentationConverters[Instrumentation.ClarityDuplicated] = clarityDuplicatedFromArray;
+instrumentationConverters[Instrumentation.ClarityActivateError] = clarityActivateErrorFromArray;
+instrumentationConverters[Instrumentation.ShadowDomInconsistent] = inconsistentShadowDomFromArray;
+
+export default instrumentationConverters;
+
+function jsErrorFromArray(jsErrorData: any[]): IJsErrorEventData {
```
toby-walker: >jsErrorFromArray [](start = 9, length = 16)

It would be better to have the to/from code in one place to make it easier to check that it matches on both sides I did *not* check that this matches as I assume tests will catch errors.

converters/fromarray/instrumentation.ts
```diff
@@ -0,0 +1,90 @@
...
+instrumentationConverters[Instrumentation.XhrError] = xhrErrorFromArray;
+instrumentationConverters[Instrumentation.TotalByteLimitExceeded] = byteLimitExceededFromArray;
+instrumentationConverters[Instrumentation.Teardown] = noDataFromArray;
+instrumentationConverters[Instrumentation.PerformanceStateError] = noDataFromArray;
+instrumentationConverters[Instrumentation.ClarityAssertFailed] = assertFailedFromArray;
+instrumentationConverters[Instrumentation.ClarityDuplicated] = clarityDuplicatedFromArray;
+instrumentationConverters[Instrumentation.ClarityActivateError] = clarityActivateErrorFromArray;
+instrumentationConverters[Instrumentation.ShadowDomInconsistent] = inconsistentShadowDomFromArray;
+
+export default instrumentationConverters;
```
toby-walker: >instrumentationConverters [](start = 15, length = 25)

How do you encode the schema version and handle schema changes?

converters/fromarray/layout.ts
```diff
@@ -0,0 +1,153 @@
+let layoutConverters = [];
+layoutConverters[Action.Insert] = insertFromArray;
+layoutConverters[Action.Remove] = removeFromArray;
+layoutConverters[Action.Move] = moveFromArray;
+layoutConverters[Action.AttributeUpdate] = attributeUpdateFromArray;
+layoutConverters[Action.CharacterDataUpdate] = cdataUpdateFromArray;
+layoutConverters[Action.Scroll] = scrollFromArray;
+layoutConverters[Action.Input] = inputFromArray;
```
toby-walker: Could we do something more generic that would just send the changed fields and a bit mas that encodes what those fields are? Seems simpler and would have less code to maintain.

converters/fromarray/pointer.ts
```diff
@@ -0,0 +1,22 @@
+let pointerConverters = [];
+pointerConverters[PointerEventType.Pointer] = pointerFromArray;
+
+export default pointerConverters;
+
+function pointerFromArray(pointerState: any[]): IPointerEventData {
```
toby-walker: >IPointerEventData [](start = 48, length = 17)

Can we optimize pointer to only send updates? At least event and position only changes are common.

declarations/clarity.d.ts
```diff
@@ -86,20 +104,30 @@ interface IImpressionMetadata {
...
+  origin: number;
+  type: number;
+  data: any;
   time?: number;
 }
 
-interface IEvent extends IEventData {
+interface IEvent extends IEventInfo {
   id: number; /* Event ID */
   time: number; /* Time relative to page start */
```
toby-walker: > time: number; /* Time relative to page start */ [](start = 1, length = 48)

Already included in IEVentInfo

clarity.d.ts
```diff
@@ -86,20 +104,30 @@ interface IImpressionMetadata {
...
+  number, // id
+  Origin, // origin
+  number, // type
+  number, // time
+  any[]   // data, converted to a value array
+];
+
 interface IDroppedPayloadInfo {
   payload: string;
-  xhrErrorState: IXhrErrorEventState;
```
toby-walker: >xhrErrorState [](start = 2, length = 13)

Why change this?

src/compressionworker.ts
```diff
@@ -57,7 +57,11 @@ function workerContext() {
     // Append new event to the next batch
     nextBatchEvents.push(event);
     nextBatchBytes += eventStr.length;
-    nextBatchIsSingleXhrErrorEvent = (nextBatchEvents.length === 1 && event.state && event.state.type === Instrumentation.XhrError);
+
+    let origin = event[1];
+    let type = event[2];
```
toby-walker: Why are type and origin the [1] and [2] elements? Obscure.

converters/fromarray.ts
```diff
@@ -0,0 +1,115 @@
...
+  } else {
+    schemas.addSchema(schema);
+  }
+
+  let data = dataFromArray(dataArray, schema as any[]);
+  let event: IEvent = { origin, type, id, time, data };
+  return event;
+}
+
+export function eventsFromDiscoverArray(
```
sarveshnagpal: >eventsFromDiscoverArray [](start = 16, length = 23)

Why do we handle discover separately in this new world of generic compression? Shouldn't it automatically be taken care of as it we process different schemas?

converters/fromarray.ts
```diff
@@ -0,0 +1,115 @@
...
+    let nextChildEvents = eventsFromDiscoverArray(id, time, children[i], index, thisEvent.data.state, previousChildState);
+    previousChildState = nextChildEvents[0].data.state;
+    index += nextChildEvents.length;
+    events = events.concat(nextChildEvents);
+  }
+
+  return events;
+}
+
+function dataFromArray(dataArray: any[], schema: any[]): any {
```
sarveshnagpal: >dataFromArray [](start = 9, length = 13)

My expectation (unless I'm missing something) is that this is the function that will take care of all cases (discover or not).

converters/schema.ts
```diff
@@ -0,0 +1,75 @@
...
+export class SchemaManager {
+
+  private schemas = {};
+
+  public getSchema(schemaHash: string): ClarityDataSchema {
+    return this.schemas[schemaHash];
+  }
+
+  public addSchema(schema: ClarityDataSchema): boolean {
+    let schemaHash = hashCode(JSON.stringify(schema));
```
sarveshnagpal: >hashCode [](start = 21, length = 8)

Why do we need hashcode for schema, we can probably prevent this indirection by storing schema string itself as key. E.g. "index,layout.x,layout.y,..." is unique in itself.

converters/schema.ts
```diff
@@ -0,0 +1,75 @@
...
+      }
+    }
+    return schema;
+  }
+
+  public getSchemaHashcode(schema: ClarityDataSchema) {
+    return hashCode(JSON.stringify(schema));
+  }
+
+  private createNestedSchema(data: object | any[], name?: string): ClarityDataSchema {
```
sarveshnagpal: >createNestedSchema [](start = 10, length = 18)

Add some comments which include the examples of schemas that it will produce. It will make it easier for future developers to maintain this code.

converters/toarray.ts
```diff
@@ -0,0 +1,82 @@
...
+  let schema = schemas.createSchema(event.data);
+  let newSchema = schemas.addSchema(schema);
+  let dataArray = dataToArray(event.data);
+  let schemaPayload = newSchema ? schema : schemas.getSchemaHashcode(schema);
+  let array = [event.id, event.origin, event.type, event.time, dataArray, schemaPayload] as IEventArray;
+  return array;
+}
+
+// When we discover the entire document or a subtree for the first time, we can structure our payload in a way
+// such that some information can be inferred from the structure, as opposed to being written explicitly and this
```
sarveshnagpal: >some information can be inferred from the structure [](start = 13, length = 51)

Let's not add this complexity since we are now dealing with redundant data pretty well in your code flow. This is an optimization we can revisit. I know this is where we started, but for the sake of simplicity, let's avoid this extra complexity.

declarations/clarity.d.ts
```diff
@@ -98,20 +116,31 @@ interface IImpressionMetadata {
...
+  number, // type
+  number, // time
+  any[],  // data, converted to a value array
+  any[] | string   // data schema, if it's a new one, otherwise - schema hashcode
+];
+
 interface IDroppedPayloadInfo {
   payload: string;
-  xhrErrorState: IXhrErrorEventState;
+  xhrError: IXhrErrorEventData;
```
sarveshnagpal: >IXhrErrorEventData [](start = 12, length = 18)

Why the rename? Curious...

declarations/test.d.ts
```diff
@@ -0,0 +1,5 @@
+declare const enum TestConstants {
```
sarveshnagpal: >TestConstants [](start = 19, length = 13)

Can we keep this enum inside the test file and not add another file if all we are going to have is this enum? I think separate of declaration for test here might be an overkill.

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "An analytics library that uses web page interactions to generate aggregated insights",
   "author": "Microsoft Corp.",
   "license": "MIT",
-  "main": "build/clarity.js",
+  "main": "build/src/module.js",
```
sarveshnagpal: >src/module.js [](start = 17, length = 13)

Have you checked if it doesn't break rest of the packages?

src/utils.ts
```diff
@@ -99,3 +98,15 @@ export function debug(text) {
     console.log(text);
   }
 }
+
+export function hashCode(str: string) {
```
sarveshnagpal: >hashCode [](start = 16, length = 8)

See my comment in schema.ts to see if it's necessary?

src/plugins/layout.ts
```diff
@@ -73,201 +68,153 @@ export default class Layout implements IPlugin {
     }
   }
 
-  // Recording full layouts of all elements on the page at once is an expensive operation
-  // and can impact user's experience by hanging the page due to occupying the thread for too long
-  // To avoid this, we only assign indices to all elements and build a ShadowDom with dummy layouts
-  // just to have a valid DOM skeleton. After that, we can come back to dummy layouts and populate
-  // them with real data asynchronously (if it takes too long to do at once) by yielding a thread
-  // and returning to it later through a set timeout
   private discoverDom() {
```
sarveshnagpal: >discoverDom [](start = 10, length = 11)

I believe all these changes are to handle discover phase different from rest of the updates. I think we can avoid that and keep the scope of the changes small. We can discuss more offline and finalize this decision.

src/plugins/layout.ts
```diff
@@ -73,201 +68,153 @@ export default class Layout implements IPlugin {
...
-        layoutState.index = eventInfo.index;
-        layoutState.action = Action.Remove;
+        event = EventProvider.createRemove(index, this.mutationSequence);
         break;
       case Action.Move:
-        layoutState.action = Action.Move;
+        let movedNode = eventInfo.node;
+        event = EventProvider.createMove(movedNode, this.mutationSequence);
+        break;
+      case Action.AttributeUpdate:
```
sarveshnagpal: >AttributeUpdate [](start = 18, length = 15)

Why do we need to know about this? When we didn't care about it before? Earlier we were looking for Updates - regardless of CharacterData update or Attributes? We should not have sub-event for different events, we should one format for layout updates.. everything else should be handled by compression layer (for saving on bytes).

src/plugins/layout.ts
```diff
@@ -1,43 +1,38 @@
...
-import { debug, isNumber, traverseNodeTree } from "./../utils";
+import { treeToDiscoverArray } from "../../converters/toarray";
+import { Action, CharacterDataNode, DiscoverEventType, IDiscover, IElementLayoutState, IEventInfo, IInputLayoutState, ILayoutEventData,
+   ILayoutEventInfo, ILayoutRoutineInfo, ILayoutState, IMutationRoutineInfo, InputElement, Instrumentation, IPlugin,
+   IShadowDomInconsistentEventData, IShadowDomMutationSummary, IShadowDomNode, ITextLayoutState, LayoutRoutine, NumberJson,
+   Origin} from "../../declarations/clarity";
+import { config } from "../config";
+import { addEvent, addMultipleEvents, bind, getTimestamp, instrument } from "../core";
+import { debug, isNumber, traverseNodeTree } from "../utils";
+import * as EventProvider from "./layout/eventprovider";
```
sarveshnagpal: What's the purpose of event provider? Are we trying to fix a bug we had? If this is for payload reduction.. we are getting into domain knowledge here, which we should avoid.

PR #45, opened 2017-12-21
Teardown from an active state only

ssivov:


clarity.d.ts
```diff
@@ -57,7 +57,9 @@ interface IConfig {
 
 declare const enum State {
   Loaded,
```
toby-walker: Can you please add the new states at the end. This will break the decoding on the client as the mapping to values will change. If you add at the end then they keep the same encoding.

src/core.ts
```diff
@@ -65,36 +66,41 @@ export function activate() {
     onActivateError(e);
     return;
   }
+
+  state = State.Activated;
```
toby-walker: What motivated this change? What bug will it fix?

src/core.ts
```diff
@@ -65,36 +66,41 @@ export function activate() {
...
 }
 
 export function teardown() {
-  for (let plugin of activePlugins) {
-    plugin.teardown();
-  }
+  if (state === State.Activating || state === State.Activated) {
```
toby-walker: >Activating [](start = 22, length = 10)

Is it OK to do this during the "Activating" state? Why? Would be good to add a comment to explain this more.

clarity.d.ts
ssivov: Done

---
In reply to: [158272085](https://github.com/Microsoft/clarity-js/pull/45#discussion_r158272085) [](ancestors = 158272085)

src/core.ts
ssivov: We can encounter errors during init phase, in which case we want to clean up.

---
In reply to: [158272666](https://github.com/Microsoft/clarity-js/pull/45#discussion_r158272666) [](ancestors = 158272666)

src/core.ts
ssivov: Before this fix attempting to teardown clarity instance, which was never activated caused a JS error.

---
In reply to: [158272415](https://github.com/Microsoft/clarity-js/pull/45#discussion_r158272415) [](ancestors = 158272415)

PR #46, opened 2018-01-05
support for a clarity customer to specify their own IG/MUIDs

ender336:


clarity.d.ts
```diff
@@ -35,6 +35,16 @@ interface IConfig {
...
 
+  // Pointer to the function which would be responsible for obtaining a
+  // page level GUID.
+  // If left unspecified, random GUID will be generated
+  igOverride?: () => string;
+
+  // Pointer to the function which would be responsible for obtaining a
+  // user level GUID.
+  // If left unspecified, the ClarityID cookie will be used or randomly generated
+  muidOverride?: () => string;
```
ssivov: >muid [](start = 2, length = 4)

We call it cid in the clarity context

clarity.d.ts
```diff
@@ -35,6 +35,16 @@ interface IConfig {
   // If left unspecified, raw payloads will be uploaded to the uploadUrl endpoint
   uploadHandler?: UploadHandler;
 
+  // Pointer to the function which would be responsible for obtaining a
+  // page level GUID.
+  // If left unspecified, random GUID will be generated
+  igOverride?: () => string;
```
ssivov: >igOverride [](start = 2, length = 10)

Use naming consistent with the variables in core - impressionId and cid

clarity.d.ts
ssivov: >muidOverride [](start = 2, length = 12)

optional: i think getCid and getImpressionId would be better names

PR #49, opened 2018-01-19
Samart/custom kvp logging

ender336:


clarity.d.ts
```diff
@@ -35,15 +35,9 @@ export interface IConfig {
...
-  // If left unspecified, random GUID will be generated
-  getImpressionId?: () => string;
-
-  // Pointer to the function which would be responsible for obtaining a
-  // user level GUID.
-  // If left unspecified, the ClarityID cookie will be used or randomly generated
-  getCid?: () => string;
+  // Pointer to the function which will be responsible for giving Clarity
+  // a dictionary of strings that the user wants logged in each Clarity payload
+  customInstrumentation?: () => Object;
```
toby-walker: >Object [](start = 32, length = 6)

Minor: Use better typing to indicate its a dictionary of string->string rather than "Object".

PR #51, opened 2018-01-22
Improving support for box model

sarveshnagpal:
1) Simplify separation of code between layout & state provider
2) Instrumented selected computed styles as part of layout state

### DIFF
--- a/clarity.d.ts
+++ b/clarity.d.ts
@@ -167,7 +167,9 @@ interface IShadowDomNode extends HTMLDivElement {
   ignore: boolean;  /* Flag to avoid sending data for that node */
 }
 
+/* Computed CSS styles associated with a layout element */
 interface ILayoutStyle {
+  visibility?: string;
   color?: string;
   backgroundColor?: string;
   backgroundImage?: string;
--- a/clarity.d.ts
+++ b/clarity.d.ts
@@ -172,6 +172,14 @@ interface IShadowDomNode extends HTMLDivElement {
   ignore: boolean;  /* Flag to avoid sending data for that node */
 }
 
+interface ILayoutStyle {
+  color?: string;
+  backgroundColor?: string;
+  backgroundImage?: string;
+  overflowX?: string;
+  overflowY?: string;
+}
+
 interface ILayoutRectangle {
   x: number; /* X coordinate of the element */
   y: number; /* Y coordinate of the element */
@@ -231,6 +239,7 @@ interface IDoctypeLayoutState extends ILayoutState {
 interface IElementLayoutState extends ILayoutState {
   attributes: IAttributes;  /* Attributes associated with an element */
   layout: ILayoutRectangle; /* Layout rectangle */
+  style: ILayoutStyle; /* Layout computed styles */
 }
 
 interface IStyleLayoutState extends IElementLayoutState {
--- a/src/plugins/layout.ts
+++ b/src/plugins/layout.ts
@@ -6,7 +6,7 @@ import { addEvent, addMultipleEvents, bind, getTimestamp, instrument } from "./.
 import { debug, isNumber, traverseNodeTree } from "./../utils";
 import { ShadowDom } from "./layout/shadowdom";
 import { createGenericLayoutState, createIgnoreLayoutState, createLayoutState } from "./layout/stateprovider";
-import { getNodeIndex, IgnoreTag, NodeIndex } from "./layout/stateprovider";
+import { getLayoutState, getNodeIndex, IgnoreTag, NodeIndex, resetStates } from "./layout/stateprovider";
 
 export default class Layout implements IPlugin {
   private eventName = "Layout";
@@ -20,7 +20,6 @@ export default class Layout implements IPlugin {
   private domDiscoverComplete: boolean;
   private lastConsistentDomJson: NumberJson;
   private firstShadowDomInconsistentEvent: IShadowDomInconsistentEventState;
-  private layoutStates: ILayoutState[];
   private originalLayouts: Array<{
     node: Node;
     layout: ILayoutState;
@@ -36,8 +35,8 @@ export default class Layout implements IPlugin {
     this.domPreDiscoverMutations = [];
     this.lastConsistentDomJson = null;
     this.firstShadowDomInconsistentEvent = null;
-    this.layoutStates = [];
     this.originalLayouts = [];
+    resetStates();
   }
 
   public activate(): void {
@@ -93,9 +92,7 @@ export default class Layout implements IPlugin {
   // Add node to the ShadowDom to store initial adjacent node info in a layout and obtain an index
   private discoverNode(node: Node) {
     this.shadowDom.insertShadowNode(node, getNodeIndex(node.parentNode), getNodeIndex(node.nextSibling));
-    let index = getNodeIndex(node);
     let layout = createGenericLayoutState(node, null);
-    this.layoutStates[index] = layout;
     this.originalLayouts.push({
       node,
       layout
@@ -118,21 +115,21 @@ export default class Layout implements IPlugin {
     while (this.originalLayouts.length > 0 && getTimestamp(true) < yieldTime) {
       let originalLayout = this.originalLayouts.shift();
       let originalLayoutState = originalLayout.layout;
-      let currentLayoutState = createLayoutState(originalLayout.node, this.shadowDom);
-
-      currentLayoutState.index = originalLayout.layout.index;
+      let currentLayoutState = createLayoutState(originalLayout.node);
       currentLayoutState.parent = originalLayoutState.parent;
       currentLayoutState.previous = originalLayoutState.previous;
       currentLayoutState.next = originalLayoutState.next;
       currentLayoutState.source = Source.Discover;
       currentLayoutState.action = Action.Insert;
 
+      // Watch for any input changes or scroll events within the element
+      this.watch(originalLayout.node, currentLayoutState);
+
       events.push({
         type: this.eventName,
         state: currentLayoutState,
         time
       });
-      this.layoutStates[originalLayout.layout.index] = currentLayoutState;
     }
     addMultipleEvents(events);
 
@@ -162,14 +159,13 @@ export default class Layout implements IPlugin {
         type: this.eventName,
         state: eventState
       });
-      this.layoutStates[eventState.index] = eventState;
     }
     addMultipleEvents(eventsData);
   }
 
   private createEventState<T extends ILayoutEventInfo>(eventInfo: T): ILayoutState {
     let node = eventInfo.node;
-    let layoutState: ILayoutState = createLayoutState(node, this.shadowDom);
+    let layoutState: ILayoutState = createLayoutState(node);
 
     switch (eventInfo.action) {
       case Action.Insert:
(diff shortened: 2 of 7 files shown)


src/plugins/layout/shadowdom.ts
```diff
@@ -36,7 +36,7 @@ export class ShadowDom {
     let shadowNode = this.doc.createElement("div") as IShadowDomNode;
     shadowNode.id = "" + index;
     shadowNode.node = node;
-    shadowNode.ignore = shouldIgnoreNode(node, this);
+    shadowNode.ignore = shouldIgnoreNode(node) || (parent && parent.ignore && parent.node !== document);
```
toby-walker: > || (parent && parent.ignore && parent.node !== document); [](start = 46, length = 58)

Move this condition to function.

clarity.d.ts
```diff
@@ -172,6 +172,14 @@ interface IShadowDomNode extends HTMLDivElement {
   ignore: boolean;  /* Flag to avoid sending data for that node */
 }
 
+interface ILayoutStyle {
+  color?: string;
+  backgroundColor?: string;
+  backgroundImage?: string;
+  overflowX?: string;
```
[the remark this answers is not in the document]
sarveshnagpal: Added comment for clarification.

src/plugins/layout/shadowdom.ts
sarveshnagpal: I moved it to a separate function. I want to avoid state provider knowing about shadowdom to simplify things here and check for parent needs to lookup shadowdom, which I am trying to avoid.

src/plugins/layout.ts
```diff
@@ -235,40 +231,42 @@ export default class Layout implements IPlugin {
...
-
-      // Update the reference of layouts object to current state
-      if (recordEvent) {
-        this.layoutStates[index] = newLayoutState;
-        addEvent({type: this.eventName, state: newLayoutState});
-      }
     }
   }
 
   private checkDistance(stateOne: IElementLayoutState, stateTwo: IElementLayoutState) {
```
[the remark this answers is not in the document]
sarveshnagpal: You are right and I removed checkDistance - we basically don't need to copy over state into a new variable. We can deal with it in a more straight forward way. I will kill checkDistance implementation.

src/plugins/layout/stateprovider.ts
```diff
@@ -59,12 +73,50 @@ export function createDoctypeLayoutState(doctypeNode: DocumentType): IDoctypeLay
...
   let elementState = createGenericLayoutState(element, tagName) as IElementLayoutState;
-  if (tagName === ScriptTag || tagName === MetaTag) {
+  if (tagName === Token.Script || tagName === Token.Meta) {
     elementState.tag = IgnoreTag;
     return elementState;
   }
 
+  elementState.attributes = getAttributes(element);
+  elementState.layout = getLayout(element);
+  elementState.style = elementState.layout ? getStyles(element) : null;
```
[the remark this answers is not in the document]
sarveshnagpal: Thanks, added comments.

src/plugins/layout/stateprovider.ts
```diff
@@ -82,42 +134,49 @@ export function createElementLayoutState(element: Element): IElementLayoutState
...
 
-    elementState.layout = {
-      x: Math.round(rect.left),
-      y: Math.round(rect.top),
-      width: Math.round(rect.width),
-      height: Math.round(rect.height)
-    };
+    // Send computed styles, if relevant, back to server
+    if (match(computed[Token.OverflowX], ["auto", "scroll", "hidden"])) {
+      style[Token.OverflowX] = computed[Token.OverflowX];
```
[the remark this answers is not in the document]
sarveshnagpal: Done

src/plugins/layout/stateprovider.ts
```diff
@@ -82,42 +134,49 @@ export function createElementLayoutState(element: Element): IElementLayoutState
...
-  // In IE, calling getBoundingClientRect on a node that is disconnected
-  // from a DOM tree, sometimes results in a 'Unspecified Error'
-  // Wrapping this in try/catch is faster than checking whether element is connected to DOM
-  let rect = null;
-  try {
-    rect = element.getBoundingClientRect();
-  } catch (e) {
-    // Ignore
-  }
+function getStyles(element) {
```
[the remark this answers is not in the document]
sarveshnagpal: Good point, I will add.

src/plugins/layout/stateprovider.ts
```diff
@@ -82,42 +134,49 @@ export function createElementLayoutState(element: Element): IElementLayoutState
...
+    }
+
+    if (computed[Token.Color] !== defaultStyles[Token.Color]) {
+      style[Token.Color] = computed[Token.Color];
+    }
+
+    return Object.keys(style).length > 0 ? style : null;
+}
+
+function match(variable, values) {
```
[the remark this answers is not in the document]
sarveshnagpal: Done, thanks.

src/plugins/layout/stateprovider.ts
```diff
@@ -187,13 +247,16 @@ export function shouldIgnoreNode(node: Node, shadowDom: ShadowDom): boolean {
   // Ignore subtrees of ignored nodes (e.g. text with a <script> parent)
   if (!ignore) {
     let parentIndex = getNodeIndex(node.parentNode);
-    if (parentIndex !== null) {
-      let parentShadowNode = shadowDom.getShadowNode(parentIndex);
-      assert(!!parentShadowNode, "shouldIgnoreNode", "parentShadowNode is missing");
-      if (parentShadowNode && parentShadowNode.ignore && parentShadowNode.node !== document) {
+    if (parentIndex !== null && parentIndex > 0) {
```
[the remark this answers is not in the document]
sarveshnagpal: Yes it can be... That will be a document node, which is what we were checking before as: && parentShadowNode.node !== document. I will add the comment.

src/plugins/layout/stateprovider.ts
```diff
@@ -1,24 +1,42 @@
...
 export const DoctypeTag = "*DOC*";
 export const TextTag = "*TXT*";
 export const IgnoreTag = "*IGNORE*";
-const MetaTag = "META";
-const ScriptTag = "SCRIPT";
+
+enum Tags {
+  Meta = "META",
+  Script = "SCRIPT"
+}
```
ssivov: Move the other tags into this enum too

src/plugins/layout.ts
```diff
@@ -162,14 +159,13 @@ export default class Layout implements IPlugin {
         type: this.eventName,
         state: eventState
       });
-      this.layoutStates[eventState.index] = eventState;
```
ssivov: Why remove this ?

src/plugins/layout/shadowdom.ts
```diff
@@ -276,6 +276,10 @@ export class ShadowDom {
     }
   }
 
+  private shouldIgnoreParent(parent) {
```
ssivov: Please add typings

src/plugins/layout/stateprovider.ts
```diff
@@ -94,30 +112,83 @@ export function createElementLayoutState(element: Element): IElementLayoutState
...
+
+    if (computed[Styles.Color] !== defaultColor) {
+      style[Styles.Color] = computed[Styles.Color];
+    }
+
+    return Object.keys(style).length > 0 ? style : null;
+}
+
+function match(variable: string, values: string[]): boolean {
+  for (let i = 0; i < values.length; i++) {
```
ssivov: values.indexOf(variable) > -1

src/plugins/layout/stateprovider.ts
```diff
@@ -94,30 +112,83 @@ export function createElementLayoutState(element: Element): IElementLayoutState
...
+  return stateAttributes;
+}
 
-  return elementState;
+function getStyles(element) {
+    let computed = window.getComputedStyle(element);
+    let style = {};
+
+    if (defaultColor.length === 0) {
+      defaultColor = computed[Styles.Color];
```
ssivov: What is default color here and why do we only set it for, what seems to be, only the first element ?

src/plugins/layout/stateprovider.ts
```diff
@@ -188,13 +260,18 @@ export function shouldIgnoreNode(node: Node, shadowDom: ShadowDom): boolean {
   // Ignore subtrees of ignored nodes (e.g. text with a <script> parent)
   if (!ignore) {
     let parentIndex = getNodeIndex(node.parentNode);
-    if (parentIndex !== null) {
-      let parentShadowNode = shadowDom.getShadowNode(parentIndex);
-      assert(!!parentShadowNode, "shouldIgnoreNode", "parentShadowNode is missing");
-      if (parentShadowNode && parentShadowNode.ignore && parentShadowNode.node !== document) {
```
ssivov: Did you lose the parentShadowNode.node !== document check intentionally ?

src/plugins/layout/stateprovider.ts
```diff
@@ -151,25 +222,26 @@ export function createIgnoreLayoutState(node: Node): IIgnoreLayoutState {
...
     parent: getNodeIndex(node.parentNode),
     previous: getNodeIndex(node.previousSibling),
     next: getNodeIndex(node.nextSibling),
     source: null,
     action: null,
     tag
   };
-  return layoutState;
+
+  return layoutStates[layoutIndex];
```
ssivov: Since you made a function for it, it would make sense to use getLayoutState(layoutIndex) here

src/plugins/layout/stateprovider.ts
sarveshnagpal: Done, thanks.

src/plugins/layout.ts
sarveshnagpal: Because we are caching states in state provider, and not layout. The benefit if we can walk the relationship between states (parent, etc.) without involving an external module like Layout. We can discuss this more offline if you would like.

src/plugins/layout/shadowdom.ts
sarveshnagpal: Done, thanks.

src/plugins/layout/stateprovider.ts
sarveshnagpal: Kept the function but replaced the body with your suggestion. I forgot we don't have to worry about older browsers here.

src/plugins/layout/stateprovider.ts
sarveshnagpal: Yeah... because "color" property is inherited. Usually you gather default styles by adding a dummy node, attach it to dom, enumerate computed styles, and dettach. In our case we will cause mutations which we don't want. If we monitor color for whatever is the first element, in practice BODY, we just ensure that we not resending that same value again & again.

src/plugins/layout/stateprovider.ts
sarveshnagpal: No, if you look at the addition, you will notice I've:
if (parentIndex !== null && parentIndex > 0) and also a comment:
// Check if parent is not null or not document node (parentIndex === 0)

src/plugins/layout/stateprovider.ts
sarveshnagpal: I'm not sure if I understand what you are referring to here. This is a set, not a get here.

PR #52, opened 2018-01-23
Adding data converters to crunch data over the network

ssivov:


clarity.d.ts
```diff
@@ -125,6 +132,15 @@ interface IBindingContainer {
 export type UploadCallback = (status: number) => void;
 export type UploadHandler = (payload: string, onSuccess?: UploadCallback, onFailure?: UploadCallback) => void;
 
+// IEvent object converted to a value array representation
+type IEventArray = [
+  number, // id
+  string, // type
+  number, // time
+  any[],  // state
+  ClarityDataSchema   // data schema, if it's a new one, otherwise - schema hashcode
```
toby-walker: >ClarityDataSchema [](start = 2, length = 17)

Where is this defined?

clarity.d.ts
```diff
@@ -443,8 +460,12 @@ interface IPerformanceResourceTiming {
...
 /* ##################################### */
 
-export function start(config?: IConfig): void;
-export function stop(): void;
+declare const enum ObjectType {
+  Object,
+  Array
+}
+
+type ClarityDataSchema = null | string | any[];
```
toby-walker: >null | string | any[]; [](start = 25, length = 22)

What do these types represent? Can you add some comment about how data schema information is represented?

converters/fromarray.ts
```diff
@@ -0,0 +1,59 @@
+import {  IEvent, IEventArray, ILayoutState } from "../clarity";
+import { Action, ObjectType } from "../clarity";
+import schemas from "./schema";
+
+export default function(eventArray: IEventArray): IEvent {
+  let id          = eventArray[0];
+  let type        = eventArray[1];
+  let time        = eventArray[2];
+  let stateArray  = eventArray[3];
+  let schema      = eventArray[4];
```
toby-walker: Can you used a destructuring assignment like

['id, type, time, stateArray, scheam] = eventArray?

?

toby-walker: You could use a struct of array rather than array of structs for encdoing. That is have something like:

{
id: [1, 5, ...]
time: [10, 20, 30]
type: ['a', 'b', ]
schema: ['schema1', ....]
state: [....]
}

Probably compresses@  better and consumes less memory.

converters/fromarray.ts
```diff
@@ -0,0 +1,59 @@
...
+    schemas.addSchema(schema);
+  }
+
+  let state = dataFromArray(stateArray, schema as any[]);
+  let event: IEvent = { id, type, time, state };
+  return event;
+}
+
+function dataFromArray(dataArray: any[], schema: any[]): any {
+  if (typeof schema === "string" || schema === null) {
```
toby-walker: >if (typeof schema === "string" || schema === null) { [](start = 2, length = 52)

When can this case happen? Add comment.

converters/fromarray.ts
```diff
@@ -0,0 +1,59 @@
...
+    return dataArray;
+  }
+
+  let data = null;
+  let dataType = null;
+  let subschemas = null;
+  if (schema.length === 2) {
+    dataType = schema[0];
+    subschemas = schema[1];
+  } else if (schema.length === 3) {
```
toby-walker: >schema.length === 3 [](start = 13, length = 19)

What's schema[0] in this case?

converters/fromarray.ts
```diff
@@ -0,0 +1,59 @@
...
+    dataType = schema[1];
+    subschemas = schema[2];
+  }
+
+  if (dataType === ObjectType.Object) {
+    data = {};
+    for (let i = 0; i < subschemas.length; i++) {
+      let nextSubschema = subschemas[i];
```
toby-walker: >nextSubschema [](start = 10, length = 13)

Minor naming suggestion: use "currentSchema". Next is a bit confusing.

converters/schema.ts
```diff
@@ -0,0 +1,75 @@
...
+export class SchemaManager {
+
+  private schemas = {};
+
+  public getSchema(schemaHash: string): ClarityDataSchema {
+    return this.schemas[schemaHash];
+  }
+
+  public addSchema(schema: ClarityDataSchema): boolean {
+    let schemaHash = hashCode(JSON.stringify(schema));
```
toby-walker: >hashCode [](start = 21, length = 8)

What if there is a collision?

converters/schema.ts
```diff
@@ -0,0 +1,75 @@
...
+    return isNew;
+  }
+
+  public createSchema(data: any, name?: string): ClarityDataSchema {
+    let schema: ClarityDataSchema = name || null;
+    if (data !== null) {
+      if (typeof data === "object") {
+        schema = this.createNestedSchema(data, name);
+      } else {
+        schema = typeof name === "string" ? name : null;
```
toby-walker: >null [](start = 51, length = 4)

This is for arrays?

converters/schema.ts
toby-walker: >hashCode(JSON.stringify(schema)); [](start = 21, length = 33)

Change to call getSchemaHashcode(). Does same thing as this line.

converters/toarray.ts
```diff
@@ -0,0 +1,47 @@
...
+
+// We serialize send a lot of JSONs with identical structures and putting all property names on the
+// wire every time can be very redundant. We can save bytes by splitting JSON into a separate nested array of
+// its property values and a separate nested array of its property names (schema). This way, each unique schema
+// has to only be sent for the first event of that type and the following events can then re-use this schema
+// on the server for reconstructing the full JSON.
+export default function(event: IEvent): IEventArray {
+  let keys = Object.keys(event);
+  let schema = schemas.createSchema(event.state);
+  let newSchema = schemas.addSchema(schema);
```
toby-walker: Combine into one function in the schemas module.

converters/toarray.ts
```diff
@@ -0,0 +1,47 @@
...
+    let keys = Object.keys(data);
+    for (let i = 0; i < keys.length; i++) {
+      let key = keys[i];
+      let val = data[key];
+      dataArray.push(dataToArray(val));
+    }
+  } else if (data.constructor === Array) {
+    for (let i = 0; i < data.length; i++) {
+      dataArray.push(dataToArray(data[i]));
+    }
```
toby-walker: Is there a native method for this.

src/utils.ts
```diff
@@ -99,3 +99,15 @@ export function debug(text) {
     console.log(text);
   }
 }
+
+export function hashCode(str: string) {
```
toby-walker: >hashCode [](start = 16, length = 8)

Very minor. Pretty weak hash function. Maybe find a better one?

clarity.d.ts
ssivov: Created a separate MD file, because the detailed explanation is perhaps too long for a code comment

converters/fromarray.ts
ssivov: Done

converters/schema.ts
ssivov: Added some comments to clarify

converters/schema.ts
ssivov: Done, thanks

converters/toarray.ts
ssivov: I'm getting two data pieces here - a schema itself and a boolean indicating whether it's a new schema or not. I don't want to return combine them into one return value as an array or some custom object.

converters/toarray.ts
ssivov: Yes, you can do...
data.forEach((dataPiece) => {
    dataArray.push(dataToArray(dataPiece));
});
... , but i doesn't really save you mich in this case. I think the for loop is more transparent.

src/utils.ts
ssivov: This is supposedly a hash function that is used in Java
http://werxltd.com/wp/2010/05/13/javascript-implementation-of-javas-string-hashcode-method/

converters/schema.ts
ssivov: I switched to hand-managed schema ids - will be shorted than hash code over the wire and no collisions possible.

ssivov:
@sarvesh is added to the review. #Closed

ssivov:
@sarveshnagpal is added to the review. #Closed

converters/schema.md
```diff
@@ -0,0 +1,60 @@
+Schema represents the structure of an arbitrary JavaScript variable. For Objects, the goal is to list its property names without property values, as an array.
```
toby-walker: Very minor: Can you wrap these lines? Very long.

PR #59, opened 2018-02-07
Expose trigger function and SchemaManager class as part of the module

ssivov:


clarity.d.ts
```diff
@@ -4,6 +4,7 @@
 
 export function start(config?: IConfig): void;
 export function stop(): void;
+export function trigger(key: string): void;
```
ender336: it's a small change, but this merits a version bump imo (so we can know which version you're referencing will contain something that can be relied on)

ssivov: The will be a version change coming from a second PR. No need to split them in two version bumps because they will be deployed together

PR #63, opened 2018-02-13
Refactoring upload logic and adding sequential uploads

ssivov:


src/core.ts
```diff
@@ -197,25 +185,19 @@ export function onWorkerMessage(evt: MessageEvent) {
...
-            onSuccess,
-            onFailure
-          };
-          pendingUploads.push(uploadInfo);
+          enqueueUpload(uploadMsg.compressedData, uploadMsg.rawData);
         } else {
-          upload(uploadMsg.compressedData, onSuccess, onFailure);
+          upload(uploadMsg.compressedData, uploadMsg.rawData);
         }
+        sequence = uploadMsg.rawData.envelope.sequenceNumber + 1;
```
ender336: >sequence = uploadMsg.rawData.envelope.sequenceNumber + 1; [](start = 8, length = 57)

what is sequence used for? I only saw it being initialized and set here, never consumed right?

ssivov: It keeps track of what is the sequence number of the next batch, so that when teardown happens and compression worker shuts down, core can upload remaining events itself, wrapped in an envelope with a correct sequence number. It is used in the "uploadPendingEvents" function on line 233.

---
In reply to: [167967974](https://github.com/Microsoft/clarity-js/pull/63#discussion_r167967974) [](ancestors = 167967974)

src/upload.ts
```diff
@@ -0,0 +1,150 @@
...
+// Storage for payloads that are compressed and are ready to be sent, but are waiting for Clarity trigger.
+// Once trigger is fired, all payloads from this queue will be sent in the order they were generated.
+let uploadQueue: IUploadInfo[];
+
+// Map from a sequence number to an UploadInfo object for payloads that were attempted to be sent, but were not succesfully delivered yet
+let uploadInfos: { [key: number]: IUploadInfo };
+
+let reUploadQueue: number[];
+
+export function upload(compressed: string, raw: IPayload, onSuccessExtra?: UploadCallback, onFailureExtra?: UploadCallback) {
```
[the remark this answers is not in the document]
ssivov: Well, like the name suggests, one will be executed on on success and one on failure. When uploading multiple payloads sequentially, we don't really care whether previous XHR succeeded or failed, we just want it to be completed, so both onSuccess and onFailure handlers are "uploadNext", but only one of them will actually get executed.

---
In reply to: [167975763](https://github.com/Microsoft/clarity-js/pull/63#discussion_r167975763) [](ancestors = 167975763)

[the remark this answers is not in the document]
ssivov: Btw not thrilled myself about the "extra" naming, let me know if you have suggestions :P They idea is that we are doing some default onSuccess/onFailure bookkeeping and then a caller has an option to pass his own additional callbacks that would be invoked as well.

---
In reply to: [167977277](https://github.com/Microsoft/clarity-js/pull/63#discussion_r167977277) [](ancestors = 167977277,167975763)

src/upload.ts
```diff
@@ -0,0 +1,150 @@
...
+}
+
+function onFailureDefault(sequenceNumber: number, status: number) {
+  debug(`FAILED: Delivery failed for batch #${sequenceNumber} Status: ${status}`);
+  let uploadInfo = uploadInfos[sequenceNumber];
+  uploadInfo.failureCount++;
+  logXhrError(sequenceNumber, status, uploadInfo);
+  if (uploadInfo.failureCount <= config.reUploadLimit) {
+    reUploadQueue.push(sequenceNumber);
+  }
```
ender336: I don't think it makes sense to continue uploading if we exceed the failure limit, best case scenario we are going to have a hole in the middle of the impression and a playback that will either not render or will be messed up

src/upload.ts
```diff
@@ -0,0 +1,150 @@
...
+    let totalByteLimitExceededEventState: ITotalByteLimitExceededEventState = {
+      type: Instrumentation.TotalByteLimitExceeded,
+      bytes: sentBytesCount
+    };
+    instrument(totalByteLimitExceededEventState);
+    teardown();
+    return;
+  }
+
+  reUpload();
```
ender336: >reUpload(); [](start = 2, length = 11)

why don't you have the reUpload() call in onFailure instead? Presumably in most cases this is going to be a noop from the success path and the failure path is really when you'd be interested in retrying right?

src/upload.ts
```diff
@@ -0,0 +1,150 @@
...
+    compressedLength: uploadInfo.compressed.length,
+    rawLength: JSON.stringify(uploadInfo.raw).length,
+    firstEventId: getEventId(events[0]),
+    lastEventId: getEventId(events[events.length - 1]),
+    attemptNumber: uploadInfo.failureCount
+  };
+  instrument(xhrErrorState);
+}
+
+function reUpload() {
```
ender336: >reUpload [](start = 9, length = 8)

rename to indicate this is for retrying failed uploads

src/upload.ts
```diff
@@ -0,0 +1,150 @@
...
+export function enqueueUpload(compressed: string, raw: IPayload) {
+  let payloadInfo: IUploadInfo = { compressed, raw, failureCount: 0 };
+  uploadQueue.push(payloadInfo);
+}
+
+export function flushUploadQueue() {
+  uploadSequentially(uploadQueue);
+}
+
+export function uploadSequentially(uploads: IUploadInfo[]) {
```
ender336: >uploadSequentially [](start = 16, length = 18)

this is potentially a naming nit, but with the way you handle errors you aren't actually going in sequence - if 2 fails but passes on retry you'd upload:

1-3-2

If the order of uploads isn't important, then this should just be renamed to indicate it is serial rather than sequential.

ender336: >export [](start = 0, length = 6)

why is this exported? Only flushUploadQueue likely should be

src/upload.ts
```diff
@@ -0,0 +1,150 @@
...
+  if (uploads.length > 0) {
+    let nextUpload = uploads.shift();
+    let uploadNext = (status: number) => {
+      uploadSequentially(uploads);
+    };
+    upload(nextUpload.compressed, nextUpload.raw, uploadNext, uploadNext);
```
ender336: >upload(nextUpload.compressed, nextUpload.raw, uploadNext, uploadNext); [](start = 4, length = 70)

the naming here makes it difficult to parse - how about keeping upload as a verb and using payload (or some other noun) to disambiguate. As it stands now you check an array of uploads then call upload on the nextUpload with an uploadNext callback, way too much of the same word being used to mean different things

src/upload.ts
ssivov: I had similar thoughts, but concluded there's still merit for analysis. We might still be able to extract things like the length of impression, instrumentation events, error data etc even if visualization is broken.

---
In reply to: [168000764](https://github.com/Microsoft/clarity-js/pull/63#discussion_r168000764) [](ancestors = 168000764)

src/upload.ts
ssivov: Presumably, if an xhr failed, there is something wrong with the connection. When the next xhr is successful, we know that network is OK and we can now attempt reuploading the failed batch.

---
In reply to: [168000770](https://github.com/Microsoft/clarity-js/pull/63#discussion_r168000770) [](ancestors = 168000770)

src/upload.ts
ssivov: I'm also nitty, yet so bad with variable names. I'll rename to uploadMultipleBatches to avoid confusion

---
In reply to: [168000793](https://github.com/Microsoft/clarity-js/pull/63#discussion_r168000793) [](ancestors = 168000793)

ssivov: You're right

---
In reply to: [168001037](https://github.com/Microsoft/clarity-js/pull/63#discussion_r168001037) [](ancestors = 168001037)

src/upload.ts
ssivov: How about retryFailedUploads ?

---
In reply to: [168000780](https://github.com/Microsoft/clarity-js/pull/63#discussion_r168000780) [](ancestors = 168000780)

src/upload.ts
ssivov: Renamed

---
In reply to: [168001870](https://github.com/Microsoft/clarity-js/pull/63#discussion_r168001870) [](ancestors = 168001870)

PR #66, opened 2018-02-21
Stopped uploading pending events when Clarity is never triggered

ssivov:


src/core.ts
```diff
@@ -222,6 +222,10 @@ function getPageContextBasedTimestamp(): number {
 }
 
 function uploadPendingEvents() {
+  // If queueUploads === true, then Clarity was never trigger and we don't want to upload any data
```
sarveshnagpal: >trigger [](start = 54, length = 7)

trigger -> triggered

Issue #64, opened 2018-02-13
Convert event types to enum


ssivov:
Currently we use strings to identify types

PR #70, opened 2018-03-01
Converting IEvent 'type' property from string to enum

ssivov:


ssivov:
Addressing issue:
https://github.com/Microsoft/clarity-js/issues/64

clarity.d.ts
```diff
@@ -81,6 +81,16 @@ declare const enum State {
   Unloading
 }
 
+declare const enum EventType {
+  /* 0 */ Discover,
+  /* 1 */ Instrumentation,
```
toby-walker: This is a breaking change for data consumption. Is the goal to save space?

clarity.d.ts
```diff
@@ -384,7 +394,8 @@ declare const enum Instrumentation {
   ClarityDuplicated,
   ShadowDomInconsistent,
   ClarityActivateError,
-  Trigger
+  Trigger,
+  PerformanceStateError
```
toby-walker: >PerformanceStateError [](start = 2, length = 21)

What does this mean? not obvious from name.

src/core.ts
```diff
@@ -11,7 +12,6 @@ import { debug, getCookie, getEventId, guid, isNumber, mapProperties, setCookie
 
 export const version = "0.1.31";
```
toby-walker: Please bump this.

test/compressionworker.ts
```diff
@@ -32,10 +32,12 @@ describe("Compression Worker Tests", () => {
 
   it("validates that events are batched for upload correctly when total length is below the limit", (done: DoneFn) => {
     let worker = createTestWorker();
-    let firstMockEventName = "FirstMockEvent";
-    let secondMockEventName = "SecondMockEvent";
-    let firstMockEvent = getMockEvent(firstMockEventName);
-    let secondMockEvent = getMockEvent(secondMockEventName);
+    let firstMockEventType = -2;
+    let secondMockEventType = -3;
```
toby-walker: What are thiese "mock" event type here for and why do you need one for first and second?

clarity.d.ts
ssivov: It is, partially, but also makes the code more neat. There are multiple other ways to reduce payloads and most of them will be breaking changes too. Let's sync offline about how to move forward with those. I won't check in this PR for now.

---
In reply to: [171736011](https://github.com/Microsoft/clarity-js/pull/70#discussion_r171736011) [](ancestors = 171736011)

clarity.d.ts
ssivov: This is the case when we detect that w3c performance entries buffer is full or has been cleared. I know it's not too obvious, but it's not a new feature and isn't a focus of this PR - i'll clean this up separately.

---
In reply to: [171736127](https://github.com/Microsoft/clarity-js/pull/70#discussion_r171736127) [](ancestors = 171736127)

ssivov: Also, there is a comment at the top of performance.ts explaining the issue

---
In reply to: [171739106](https://github.com/Microsoft/clarity-js/pull/70#discussion_r171739106) [](ancestors = 171739106,171736127)

toby-walker: Can you just add a brief comment here or refer to the appropriate location as a comment here.; 

---
In reply to: [171740258](https://github.com/Microsoft/clarity-js/pull/70#discussion_r171740258) [](ancestors = 171740258,171739106,171736127)

PR #74, opened 2018-03-15
Switch performance entry blacklist from domain-based to full URL match

ssivov:


src/plugins/performance.ts
```diff
@@ -44,8 +48,7 @@ export default class PerformanceProfiler implements IPlugin {
     this.incompleteEntryIndices = [];
 
     if (config.uploadUrl) {
```
swaathee: shouldnt the conditions be consistent? (.length > 0)

ssivov: Actually i don't need that in "activate" function at all - i didn't see that i'm already setting it in "reset" :) Thanks !

ssivov: Actually i don't need that in "activate" function at all - i didn't see that i'm already setting it in "reset" :) Thanks !

---
In reply to: [174886525](https://github.com/Microsoft/clarity-js/pull/74#discussion_r174886525) [](ancestors = 174886525)

PR #75, opened 2018-04-03
Improved text masking

ssivov:


src/plugins/layout/stateprovider.ts
```diff
@@ -138,7 +138,7 @@ function getAttributes(element) {
 
     // If we are masking text, also mask it from input boxes as well as alt description
     if (!config.showText && attributeMaskList.indexOf(attrName) >= 0) {
-      stateAttributes[attr.name] = attr.value.replace(/\S/gi, "*");
+      stateAttributes[attr.name] = attr.value.replace(/./g, "*");
```
toby-walker: Small improvement: Refactor the masking into common code so we can update/change more easily.

test/layout.ts
```diff
@@ -974,6 +976,8 @@ describe("Layout Tests", () => {
     let observer = new MutationObserver(callback);
     observer.observe(document, { childList: true, subtree: true });
 
+    config.showText = true;
```
toby-walker: >  config.showText = true; [](start = 2, length = 25)

How does the test check that masking is done? Can you add a test for the case with spaces so that we mask everything now?

PR #85, opened 2018-04-19
Send resource timings individually for better schema reuse

ssivov:


clarity.d.ts
```diff
@@ -470,7 +470,11 @@ interface IPerformanceTiming extends PerformanceTiming {
   // We send back all properties from performance.timing object
 }
 
-interface IPerformanceResourceTiming {
+interface IPerformanceTimingState {
+  timing: IPerformanceTiming;
+}
+
+interface IPerformanceResourceTimingState {
```
sarveshnagpal: >IPerformanceResourceTimingState [](start = 10, length = 31)

Chrome also sends protocol.. let's capture that too.

clarity.d.ts
```diff
@@ -470,7 +470,11 @@ interface IPerformanceTiming extends PerformanceTiming {
   // We send back all properties from performance.timing object
 }
 
-interface IPerformanceResourceTiming {
+interface IPerformanceTimingState {
```
sarveshnagpal: Call it what the spec says, IResourceTiming? PerformanceTIming is confusing because that could mean NavigationTiming, or UserTiming or PaintTiming, etc.

ssivov: Spec says PerformanceTiming
[https://developer.mozilla.org/en-US/docs/Web/API/PerformanceTiming](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceTiming)

---
In reply to: [182905491](https://github.com/Microsoft/clarity-js/pull/85#discussion_r182905491) [](ancestors = 182905491)

PR #88, opened 2018-04-20
swradhak/fixdocumentheight

swaathee:
Fix document height to take max of body height and document element height.

### DIFF
--- a/src/plugins/viewport.ts
+++ b/src/plugins/viewport.ts
@@ -40,13 +40,24 @@ export default class Viewport implements IPlugin {
       },
       document: {
         width: body ? body.clientWidth : null,
-        height: body ? body.clientHeight : null
+        height: getDocumentHeight(body, de);
       },
       dpi: "devicePixelRatio" in window ? window.devicePixelRatio : -1,
       visibility: "visibilityState" in document ? document.visibilityState : "default",
       event: type
     };
     return viewport;
+  }
+
+  private getDocumentHeight(body, documentElement) {
+     let bodyClientHeight = body ? body.clientHeight : null;
+     let bodyScrollHeight = body ? body.scrollHeight : null;
+     let bodyOffsetHeight = body ? body.offsetHeight : null;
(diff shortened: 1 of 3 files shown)


src/plugins/viewport.ts
```diff
@@ -49,6 +49,18 @@ export default class Viewport implements IPlugin {
     return viewport;
   }
 
+  private getDocumentHeight(body, documentElement): number {
```
ssivov: >body, documentElement [](start = 28, length = 21)

These shouldn't be parameters

ssivov:
>     let body = document.body;

You can declare these as class properties and re-use

---
Refers to: src/plugins/viewport.ts:33 in cc112d1. [](commit_id = cc112d1933f6a17360c184d93272bb10484874b9, deletion_comment = False)

src/plugins/viewport.ts
ssivov: >getDocumentHeight [](start = 10, length = 17)

Please add a short explanation why this extra complexity is needed vs. what we had before

src/plugins/viewport.ts
```diff
@@ -49,6 +49,18 @@ export default class Viewport implements IPlugin {
...
+     let bodyClientHeight = body ? body.clientHeight : null;
+     let bodyScrollHeight = body ? body.scrollHeight : null;
+     let bodyOffsetHeight = body ? body.offsetHeight : null;
+     let documentClientHeight = documentElement ? documentElement.clientHeight : null;
+     let documentScrollHeight = documentElement ? documentElement.scrollHeight : null;
+     let documentOffsetHeight = documentElement ? documentElement.offsetHeight : null;
+     let documentHeight = Math.max(bodyClientHeight, bodyScrollHeight, bodyOffsetHeight,
```
ssivov: >Math.max [](start = 26, length = 8)

Can any of these have negative values? If some of them are negative and some are 'null', then Math.max(-5, null) = 0

ssivov:
Please bump version

PR #90, opened 2018-04-25
Use efficient payload format for initial DOM discovery

ssivov:


ssivov:
Will revisit this later

PR #92, opened 2018-05-09
Clean up unused ShadowDom node references / Simplify mutation processing

ssivov:


ssivov:
@sarveshn is added to the review. #Closed

src/plugins/layout.ts
```diff
@@ -220,8 +170,7 @@ export default class Layout implements IPlugin {
       this.checkConsistency(actionInfo);
 
       if (this.allowMutation()) {
-        let events = this.processMutations(summary, time);
-        this.processMultipleNodeEvents(events);
+        this.processMutations(summary, time);
       } else {
         debug(`>>> ShadowDom doesn't match PageDOM after mutation batch #${this.mutationSequence}!`);
```
toby-walker: >debug [](start = 8, length = 5)

Didn't we use to send an instrumentation event in this case? ShadowDoimINcionsitent...

src/plugins/layout/shadowdom.ts
```diff
@@ -252,6 +252,8 @@ export class ShadowDom {
     let removedNodes = this.removedNodes.childNodes;
     for (let i = 0; i < removedNodes.length; i++) {
       traverseNodeTree(removedNodes[i], (shadowNode: IShadowDomNode) => {
+        let index = getNodeIndex(shadowNode.node);
+        delete this.nodeMap[index];
         if (this.hasClass(shadowNode, NewNodeClassName)) {
           delete shadowNode.node[NodeIndex];
```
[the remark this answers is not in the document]
ssivov: yeah, 'index' is the actual index of the element, and "NodeIndex" is a string property name == 'clarity-index'. So index == node[NodeIndex]

---
In reply to: [187409885](https://github.com/Microsoft/clarity-js/pull/92#discussion_r187409885) [](ancestors = 187409885)

src/plugins/layout.ts
ssivov: We do, but it's inside the 'allowMutation() -> checkConsistency()' function below. I should probably move this debug statement there as well. 

---
In reply to: [187409318](https://github.com/Microsoft/clarity-js/pull/92#discussion_r187409318) [](ancestors = 187409318)

src/plugins/layout.ts
```diff
@@ -91,54 +89,6 @@ export default class Layout implements IPlugin {
...
-        type: this.eventName,
-        state: eventState
-      });
-    }
-    addMultipleEvents(eventsData);
-  }
-
-  private createEventState<T extends ILayoutEventInfo>(eventInfo: T): ILayoutState {
-    let node = eventInfo.node;
-    let shadowNode = this.shadowDom.getShadowNode(eventInfo.index);
```
ssivov: >let shadowNode = this.shadowDom.getShadowNode(eventInfo.index); [](start = 4, length = 63)

Changes in this file are required because now that we don't preserve information about removed nodes in the shadowDom, this would cause an error when creating event state for a removed node. We need to pass detached shadowdom nodes in a summary and use them directly. Once that adjustment is made, the follow-on simplifications in this file occur naturally.

PR #97, opened 2018-07-26
ensuring our client id cookie spans multiple sessions

ender336:
Not giving an expiration for our cookies results in it only being used for the session (see https://stackoverflow.com/questions/14196671/session-only-cookies-with-javascript).

Setting our expiration date to be roughly 2 years so we can compute metrics over long timespans, though the clarityId will continue to be anonymous with no ties to identifiable information.

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -10,7 +10,7 @@ import getPlugin from "./plugins";
 import { enqueuePayload, flushPayloadQueue, resetUploads, upload } from "./upload";
 import { getCookie, getEventId, guid, isNumber, setCookie } from "./utils";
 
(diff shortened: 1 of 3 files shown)


src/core.ts
```diff
@@ -245,7 +245,8 @@ function uploadPendingEvents() {
 function init() {
   // Set ClarityID cookie, if it's not set already
   if (!getCookie(Cookie)) {
-    setCookie(Cookie, guid());
+    // setting our ClarityId cookie for 2 years
```
toby-walker: >2 years [](start = 40, length = 7)

Please see comment in email.

PR #104, opened 2018-09-06
Organizing exported type declarations for use in other npm modules

ssivov:
Providing proper npm types for our package and exposing all internal interfaces for on-demand use in other npm packages.


package.json
```diff
@@ -1,10 +1,11 @@
 {
   "name": "clarity-js",
-  "version": "0.1.44",
+  "version": "0.2.0",
```
toby-walker: Why the bump to 0.2?

ssivov: Breaking changes in the exported module structure. This would prevent somebody who's using v0.1.x from accidentally installing this version by just running "npm install"

ender336: Sorry if I missed a conversation between you and Swaathee - why do the dashboard and analytics services needs all of these changes?

---
In reply to: [215666151](https://github.com/Microsoft/clarity-js/pull/104#discussion_r215666151) [](ancestors = 215666151)

ssivov: There was no such conversation - this is my initiative. Tslint in dashboard requires typed variables in functions. Therefore, moving visualization event parsers to dashboard needs to have access to most Clarity interfaces. I don't want to add hacky "node_modules/..... /clarity.d.ts" or /// <reference path=".." /> like we have now, hence the change for proper typing setup on the package.

PR #108, opened 2018-10-26
always blacklist uploadUrl to avoid instrumenting Clarity network request (in an infinite loop)

ender336:


src/clarity.ts
```diff
@@ -8,6 +8,15 @@ export { version } from "./core";
...
     mapProperties(customConfig, null, true, config);
+    const uploadUrl = config.uploadUrl;
+    if(uploadUrl) {
+      if(!config.urlBlacklist) {
+        config.urlBlacklist = [uploadUrl]
+      } else if (config.urlBlacklist.indexOf(uploadUrl) === -1){
+          config.urlBlacklist.push(uploadUrl);
+      }
+    }
+
```
ssivov: This isn't a right place to take care of this. Take a look at performance.ts: 50

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clarity-js",
-  "version": "0.2.0",
+  "version": "0.2.1",
```
ssivov: >0.2.1 [](start = 14, length = 5)

Update version in core

ssivov: Oh, that's right, you didn't update package.json somehow. Update package.json and core.ts


---
In reply to: [228675204](https://github.com/Microsoft/clarity-js/pull/108#discussion_r228675204) [](ancestors = 228675204)

src/plugins/performance.ts
```diff
@@ -48,6 +48,7 @@ export default class PerformanceProfiler implements IPlugin {
     this.stateError = false;
     this.incompleteEntryIndices = [];
     this.urlBlacklist = config.urlBlacklist.map(this.getFullUrl);
+    this.urlBlacklist.push(this.getFullUrl(config.uploadUrl));
```
toby-walker: Add a comment about why this is needed and its importance.

PR #112, opened 2018-11-14
Ensure we track current impression id

ender336:


src/core.ts
```diff
@@ -252,9 +252,13 @@ function init() {
 
   startTime = getUnixTimestamp();
   sequence = 0;
+
+  // storing the current impressionId in a global var to determine when multiple instances of
+  // Clarity are attempting to run simultaneously
+  impressionId = guid();
```
ssivov: This check is performed in the prepare() function below (line 297)

ssivov: >impressionId = guid(); [](start = 2, length = 22)

Place this on line 252 right below where cid is set and remove this comment - it's not directly relevant here

PR #113, opened 2018-11-26
Add Mark (custom) events

ssivov:


ssivov:
@ender336 , can you please take a look?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clarity-js",
-  "version": "0.2.4",
+  "version": "0.2.5",
```
toby-walker: Out of date.

src/core.ts
```diff
@@ -10,9 +12,10 @@ import getPlugin from "./plugins";
 import { enqueuePayload, flushPayloadQueue, resetUploads, upload } from "./upload";
 import { getCookie, getEventId, guid, isNumber, setCookie } from "./utils";
 
-export const version = "0.2.4";
+export const version = "0.2.5";
 export const ClarityAttribute = "clarity-iid";
 export const InstrumentationEventName = "Instrumentation";
+const MarkEventName = "Mark";
```
toby-walker: >Mark [](start = 23, length = 4)

What does a "Mark" event mean? Not obvious from the name.

src/core.ts
```diff
@@ -175,7 +178,14 @@ export function getTimestamp(unix?: boolean, raw?: boolean) {
 
 export function instrument(eventState: IInstrumentationEventState) {
   if (config.instrument) {
-    addEvent({type: InstrumentationEventName, state: eventState});
+    addEvent({ type: InstrumentationEventName, state: eventState });
+  }
+}
+
+export function mark(key: string, info: any) {
```
toby-walker: >any [](start = 40, length = 3)

Might want to require this to be a property bag, not allow any JS type.

src/core.ts
ssivov: It is custom event for the user at any time in the page's lifecycle. The inspiration is from performance.mark() API, which lets you create custom performance entries in the browser:
[https://developer.mozilla.org/en-US/docs/Web/API/Performance/mark](https://developer.mozilla.org/en-US/docs/Web/API/Performance/mark)

---
In reply to: [250282778](https://github.com/Microsoft/clarity-js/pull/113#discussion_r250282778) [](ancestors = 250282778)

package.json
ssivov: Updated

---
In reply to: [250282609](https://github.com/Microsoft/clarity-js/pull/113#discussion_r250282609) [](ancestors = 250282609)

src/core.ts
ssivov: Since this is custom instrumentation, i don't think we can dictate what format payload should have. We can make these events searchable by keys on the backend, but it'd be up to webmaster to write/read payload later on

---
In reply to: [250283112](https://github.com/Microsoft/clarity-js/pull/113#discussion_r250283112) [](ancestors = 250283112)

ssivov:
Abandoning

PR #122, opened 2019-01-21
Force same timestamp for all discovers and same batch mutations

ssivov:
Having same timestamp on multiple events allows for easier batch processing on the backend

### DIFF
--- a/src/plugins/layout.ts
+++ b/src/plugins/layout.ts
@@ -91,13 +91,8 @@ export default class Layout implements IPlugin {
     }
   }
 
-  // Recording full layouts of all elements on the page at once is an expensive operation
-  // and can impact user's experience by hanging the page due to occupying the thread for too long
-  // To avoid this, we only assign indices to all elements and build a ShadowDom with dummy layouts
-  // just to have a valid DOM skeleton. After that, we can come back to dummy layouts and populate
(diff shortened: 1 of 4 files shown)


toby-walker:
The PR looks fine but it's missing why you are making this change. Can you explain the context.?

ssivov:
Discover and mutations are designed to be atomic operations. Giving them different timestamps creates possibilities of unexpected states in processing logic (especially visualization).

---
In reply to: [456186003](https://github.com/Microsoft/clarity-js/pull/122#issuecomment-456186003) [](ancestors = 456186003)

src/plugins/layout.ts
```diff
@@ -241,7 +243,10 @@ export default class Layout implements IPlugin {
     let moves = summary.movedNodes.map(this.processMutation.bind(this, Action.Move));
     let updates = summary.updatedNodes.map(this.processMutation.bind(this, Action.Update));
     let removes = summary.removedNodes.map(this.processMutation.bind(this, Action.Remove));
-    let all = [].concat(inserts, moves, updates, removes);
+    let all: IEventData[] = [].concat(inserts, moves, updates, removes);
+    for (let i = 0; i < all.length; i++) {
```
toby-walker: Can you add a comment here that includes your response to my question about "why this PR?" please.

ssivov: Done

PR #123, opened 2019-01-21
Split umbrella 'types' imports into distinct module imports

ssivov:
Purely housekeeping work. Having imports from particular modules rather than from a single unified 'types' module, makes reading and understand cross-modules dependencies easier.

### DIFF
--- a/src/converters/fromarray.ts
+++ b/src/converters/fromarray.ts
@@ -1,6 +1,6 @@
-import { ObjectType } from "../../types/index";
-import {  IEvent, IEventArray } from "../../types/index";
+import { IEvent, IEventArray, ObjectType } from "../../types/core";
 import { SchemaManager } from "./schema";
+
 import defaultSchemas from "./schema";
 
 export default function(eventArray: IEventArray, schemas?: SchemaManager): IEvent {
--- a/src/core.ts
+++ b/src/core.ts
@@ -1,23 +1,10 @@
 import { IAddEventMessage, ICompressedBatchMessage, ITimestampedWorkerMessage, WorkerMessageType } from "../types/compressionworker";
 import {
-  IBindingContainer,
-  IClarityFields,
-  IEnvelope,
-  IEvent,
-  IEventArray,
-  IEventBindingPair,
-  IEventData,
-  IPayload,
-  IPlugin,
-  State
+  IBindingContainer, IClarityFields, IEnvelope, IEvent, IEventArray, IEventBindingPair, IEventData, IPayload, IPlugin, State
 } from "../types/core";
 import {
-  IClarityActivateErrorState,
-  IClarityDuplicatedEventState,
-  IInstrumentationEventState,
-  IMissingFeatureEventState,
-  Instrumentation,
-  ITriggerState
+  IClarityActivateErrorState, IClarityDuplicatedEventState, IInstrumentationEventState,
+  IMissingFeatureEventState, Instrumentation, ITriggerState
 } from "../types/instrumentation";
 import { createCompressionWorker } from "./compressionworker";
 import { config } from "./config";
--- a/src/core.ts
+++ b/src/core.ts
@@ -1,15 +1,33 @@
-import { IAddEventMessage, IBindingContainer, IClarityActivateErrorState, IClarityDuplicatedEventState, IClarityFields,
-  ICompressedBatchMessage, IEnvelope, IEvent, IEventArray, IEventBindingPair, IEventData,
-  IInstrumentationEventState, IMissingFeatureEventState, Instrumentation, IPayload, IPlugin, ITimestampedWorkerMessage,
-  ITriggerState, State, WorkerMessageType } from "../types/index";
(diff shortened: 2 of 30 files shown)


src/core.ts
```diff
@@ -1,16 +1,34 @@
-import { IAddEventMessage, IBindingContainer, IClarityActivateErrorState, IClarityDuplicatedEventState, IClarityFields,
-  ICompressedBatchMessage, IEnvelope, IEvent, IEventArray, IEventBindingPair, IEventData,
-  IInstrumentationEventState, IMissingFeatureEventState, Instrumentation, IPayload, IPlugin, ITimestampedWorkerMessage,
-  ITriggerState, State, WorkerMessageType } from "../types/index";
-import compress from "./compress";
+import { IAddEventMessage, ICompressedBatchMessage, ITimestampedWorkerMessage, WorkerMessageType } from "../types/compressionworker";
+import {
+  IBindingContainer,
+  IClarityFields,
+  IEnvelope,
```
toby-walker: Minor but why are some of these multiline and not others?

src/plugins/viewport.ts
```diff
@@ -1,4 +1,5 @@
-import { IPlugin, IViewportState } from "../../types/index";
+import { IPlugin } from "../../types/core";
+import { IViewportState } from "../../types/viewport";
```
toby-walker: >./.. [](start = 33, length = 4)

Could you path aliases to make these path saner?

src/converters/fromarray.ts
```diff
@@ -1,6 +1,6 @@
-import { ObjectType } from "../../types/index";
-import {  IEvent, IEventArray } from "../../types/index";
+import { IEvent, IEventArray, ObjectType } from "../../types/core";
```
toby-walker: Where are these types/core files? Are they automatically generated? I don't see them in the PR.

src/plugins/viewport.ts
ssivov: Not yet, some infrastructure changes from other PRs are required to use path aliases

src/converters/fromarray.ts
ssivov: They all reside in the ./types folder. They aren't in the PR because i haven't touched them - just importing from different files

src/plugins/viewport.ts
ssivov: Not yet, some infrastructure changes from other PRs are required to use path aliases

---
In reply to: [250244043](https://github.com/Microsoft/clarity-js/pull/123#discussion_r250244043) [](ancestors = 250244043)

src/converters/fromarray.ts
ssivov: They all reside in the ./types folder. They aren't in the PR because i haven't touched them - just importing from different files


---
In reply to: [250244704](https://github.com/Microsoft/clarity-js/pull/123#discussion_r250244704) [](ancestors = 250244704)

src/core.ts
ssivov: Max length limitations. I turned the ones that didn't fit on a single line into multi-lines. Maybe it's not the best as far as consistency goes, i can change it to the original format like the "types/index" import that I removed above.

---
In reply to: [250243754](https://github.com/Microsoft/clarity-js/pull/123#discussion_r250243754) [](ancestors = 250243754)

ender336: I personally prefer the multiple imported elements per line rather than 1 per line, otherwise we have an entire fold on my screen essentially of just imports

---
In reply to: [250326095](https://github.com/Microsoft/clarity-js/pull/123#discussion_r250326095) [](ancestors = 250326095,250243754)

src/converters/fromarray.ts
ender336: So I thought it was weird that we had all of our typing files in the same folder and also an uber typing file that connected them all together. It looks like the standard is to keep our .d.ts files located with the code that implements them (which makes some sense to me, easy to go from the typing file to the implementation) and then our uber file would continue to grab from the different locations and export them from a single place. 

[https://github.com/Microsoft/TypeScript/issues/8305](https://github.com/Microsoft/TypeScript/issues/8305)

Thoughts?

---
In reply to: [250325003](https://github.com/Microsoft/clarity-js/pull/123#discussion_r250325003) [](ancestors = 250325003,250244704)

src/core.ts
ssivov: Yeah, ok, that makes sense

---
In reply to: [250243754](https://github.com/Microsoft/clarity-js/pull/123#discussion_r250243754) [](ancestors = 250243754)

src/converters/fromarray.ts
ssivov: Looks like there are some great ideas out there and perhaps better ways than what we are doing right now. I'm scoping it outside of this PR though, feel free to create an issue to reorganize typedefs

---
In reply to: [250692834](https://github.com/Microsoft/clarity-js/pull/123#discussion_r250692834) [](ancestors = 250692834,250325003,250244704)

PR #126, opened 2019-01-22
Use Webpack to build Clarity bundle

ssivov:


webpack/globalize.ts
```diff
@@ -0,0 +1,5 @@
+import * as clarity from "../src/clarity";
```
toby-walker: Why is this file needed for webpack? Can you add a comment to explain that?

webpack/configs/base.ts
```diff
@@ -0,0 +1,36 @@
...
+        extensions: [".ts", ".mjs", ".js", ".json"],
+        plugins: [new TsConfigPathsPlugin()]
+    },
+
+    module: {
+        rules: [
+            // All files with a '.ts' extension will be handled by 'awesome-typescript-loader'.
+            { test: /\.ts$/, loader: "awesome-typescript-loader" },
+
+            { test: /\.mjs$/, include: /node_modules/, type: "javascript/auto" },
```
toby-walker: Add a comment to this line explaining that this is for (like the lines above and below.

webpack/configs/base.ts
```diff
@@ -0,0 +1,36 @@
+import * as webpack from "webpack";
+
+import { TsConfigPathsPlugin } from "awesome-typescript-loader";
+
+const CommongConfig: webpack.Configuration = {
+
+    entry: "./webpack/globalize.ts",
+
+    output: {
+        path: `${__dirname}/../../build`
```
toby-walker: This relative path is still a bit messy.

webpack/configs/base.ts
```diff
@@ -0,0 +1,36 @@
...
+
+            { test: /\.mjs$/, include: /node_modules/, type: "javascript/auto" },
+
+            // All output '.js' files will have any sourcemaps re-processed by 'source-map-loader'.
+            { enforce: "pre", test: /\.js$/, loader: "source-map-loader" }
+        ]
+    },
+
+    performance: {
+        hints: false
```
toby-walker: What does this mean? If you can add comment to explain the webpack config entries that would be great.

webpack/configs/prod.ts
```diff
@@ -0,0 +1,34 @@
...
+const ProdConfig: webpack.Configuration = {
+
+    mode: "production",
+
+    output: {
+        filename: "clarity.min.js"
+    },
+
+    plugins: [
+        new webpack.DefinePlugin({
```
toby-walker: What does this mean and do?

webpack/configs/prod.ts
```diff
@@ -0,0 +1,34 @@
...
+    optimization: {
+        minimizer: [
+            new UglifyJsPlugin({
+                cache: false
+            })
+        ]
+    },
+
+    stats: {
+        warnings: false
```
toby-walker: Why turn of warnings?

webpack/globalize.ts
ssivov: Webpack bundle doesn't store the target module in a global variable by default like rollup used to do for us, so we need an extra wrapper module that does this manually.

---
In reply to: [250238167](https://github.com/Microsoft/clarity-js/pull/126#discussion_r250238167) [](ancestors = 250238167)

webpack/configs/prod.ts
ssivov: We don't care about build warnings in production

---
In reply to: [250239240](https://github.com/Microsoft/clarity-js/pull/126#discussion_r250239240) [](ancestors = 250239240)

webpack/configs/prod.ts
ssivov: This sets environment variable to "production". We actually don't need this in clarity-js, since it's a node feature - removing

---
In reply to: [250239148](https://github.com/Microsoft/clarity-js/pull/126#discussion_r250239148) [](ancestors = 250239148)

webpack/configs/base.ts
ssivov: Performance/Bundle size notifications from webpack. I should actually enable it as warning for development.

---
In reply to: [250238985](https://github.com/Microsoft/clarity-js/pull/126#discussion_r250238985) [](ancestors = 250238985)

webpack/globalize.ts
ssivov: Added comment as well

---
In reply to: [250333677](https://github.com/Microsoft/clarity-js/pull/126#discussion_r250333677) [](ancestors = 250333677,250238167)

webpack/configs/base.ts
ssivov: Thanks for pointing this out to me - we actually don't need this for this project, so i'm removing this as well

---
In reply to: [250238437](https://github.com/Microsoft/clarity-js/pull/126#discussion_r250238437) [](ancestors = 250238437)

webpack/configs/base.ts
ssivov: I don't think it's worth explaining every single property,because there is a Webpack Configuration documentation page out there. What i did is I placed a link to those docs at the top of each file.

---
In reply to: [250336418](https://github.com/Microsoft/clarity-js/pull/126#discussion_r250336418) [](ancestors = 250336418,250238985)

webpack/configs/base.ts
ssivov: It's actually an absolute path to "build" from the file system root. The problem is that there is no system variable that links into the project root, but only the __dirname, which points to the location of the config file, so we have to trace back a couple folders manually to get to "build".

---
In reply to: [250238812](https://github.com/Microsoft/clarity-js/pull/126#discussion_r250238812) [](ancestors = 250238812)

webpack/configs/prod.ts
```diff
@@ -0,0 +1,34 @@
...
+    optimization: {
+        minimizer: [
+            new UglifyJsPlugin({
+                cache: false
+            })
+        ]
+    },
+
+    performance: {
+        hints: false
```
toby-walker: Add comment about why this is here and only in prod.

tsconfig.json
```diff
@@ -10,7 +10,7 @@
         "declaration": false,
         "noImplicitAny": false,
         "noImplicitUseStrict": false,
-        "noUnusedLocals": true,
+        "noUnusedLocals": false,
```
ender336: >"noUnusedLocals": false, [](start = 8, length = 24)

why? if you really need an unused local (I'm assuming the window.clarity from globalize?) we should just be disabling the rule for that line.

package.json
```diff
@@ -51,15 +54,20 @@
...
+    "typescript": "^3.2.4",
+    "uglifyjs-webpack-plugin": "^2.1.1",
+    "watchify": "^3.11.0",
+    "webpack": "^4.29.0",
+    "webpack-cli": "^3.2.1",
+    "webpack-merge": "^4.2.1"
   },
   "scripts": {
     "init": "npm install && npm install -g gulp && npm install -g gulp-cli",
+    "build": "gulp build && npm run build:bundle && npm run build:bundle:dev",
```
ender336: > && npm run build:bundle && npm run build:bundle:dev [](start = 24, length = 52)

I don't think it makes sense to build both prod and dev for the default build command. A better layout here would be a separate build command from build:dev - also inlining the build:bundle calls, they are only one command. Once we separate build and build:dev they will be unnecessary

tsconfig.json
ssivov: No, this is actually for something in our 3rd party compress/decompress code, which we technically can't touch. And also, unlike tslint, you can't disable the tsconfig rule per-line.

---
In reply to: [250700312](https://github.com/Microsoft/clarity-js/pull/126#discussion_r250700312) [](ancestors = 250700312)

Issue #125, opened 2019-01-22
Test failures: Compression workers timing out


ssivov:
Compression workers started timing out in Chrome Headless. And in general, headless browser tests behavior became quite flaky. We need to fix this ASAP.

ssivov:
Compression Workers started timing out, because Chrome updated to 71, which has broken behavior for loading Blob URLs:
https://github.com/GoogleChrome/puppeteer/issues/3463

ssivov:
All tests are now fixed

PR #132, opened 2019-02-04
Make Tests Great Again

ssivov:
This PR:
  - Makes all tests pass
  - Makes all tests consistent and deterministic
  - Simplifies all tests significantly and makes them more readable/understandable
  - Builds solid api stub and mock objects tools
  - Fixes a production bug where config was not reset between consecutive clarity restarts
  - Adds few extra tests
  - Changes webpack index.ts build to production without sourcemaps (this is unrelated to tests, but slipping in this small fix)


ssivov:
Resolves #125

karma/setup/pubsub.ts
```diff
@@ -0,0 +1,103 @@
+import * as PubSub from "pubsub-js";
```
sarveshnagpal: >PubSub [](start = 12, length = 6)

All this appears to be boiler plate code. Can we not get it from npm package instead of manually copy pasting?

ssivov: I'm not sure what you mean, can you clarify ?

---
In reply to: [253637913](https://github.com/Microsoft/clarity-js/pull/132#discussion_r253637913) [](ancestors = 253637913)

Issue #82, opened 2018-04-19
Privacy: Improve privacy tests


ssivov:
Expand a set of tests ensuring we don't capture any text or images when config disallows it, or when privacy option is set on a node and its subtree. Because we value our users' privacy, our text/image masking tests must be really thorough.

ssivov:
Because of the simpler, more straightforward masking architecture, privacy testing is now more robust and is more reliable:
#82

ssivov:
With #137, everything is masked by default and there are no config options to reverse it. The only way to unmask data is to add 'data-clarity-unmask' attribute to DOM and that has its own reverse tests.

PR #137, opened 2019-02-21
Privacy: Turn ON masking as a permanent, non-configurable feature

ssivov:
The only way to unmask data is for the webmaster himself to explicitly add a "data-clarity-unmask" to the node.


ssivov:
Because of the simpler, more straightforward masking architecture, privacy testing is now more robust and is more reliable:
#82

src/plugins/layout/nodeinfo.ts
```diff
@@ -1,50 +1,22 @@
...
 
 export function createNodeInfo(node: Node, parentInfo: INodeInfo): INodeInfo {
-  let index = getNodeIndex(node);
-  let ignore = shouldIgnoreNode(node, parentInfo ? parentInfo.ignore : false);
-  let forceMask = shouldForceMaskNode(node, parentInfo ? parentInfo.forceMask : false);
-  let state: ILayoutState = createLayoutState(node, ignore, forceMask);
-  return { index, ignore, forceMask, state };
+  let ignore = shouldIgnoreNode(node, parentInfo);
+  let unmask = shouldUnmaskNode(node, parentInfo);
+  let isCss = isCssText(node);
```
toby-walker: >isCssText [](start = 14, length = 9)

Where did this go?

src/plugins/layout/nodeinfo.ts
```diff
@@ -1,50 +1,22 @@
...
-function shouldForceMaskNode(node: Node, parentForceMasked: boolean): boolean {
-  let mask = parentForceMasked;
-  if (!mask) {
-    if (node && node.nodeType === Node.ELEMENT_NODE) {
-      mask = (node as Element).getAttribute(ForceMaskAttribute) === "true";
-    }
-  }
-  return mask;
+function shouldUnmaskNode(node: Node, parentInfo: INodeInfo): boolean {
+  const parentUnmasked = parentInfo ? parentInfo.unmask : false;
```
toby-walker: >unmask [](start = 49, length = 6)

do we guarantee that the parent's state is always set correctly before here or do we need to traverse the path to the root to make sure? Traversing the path and having up until the first parent that has an explicit value set seems "safer".

toby-walker:
>     y: "pageY" in evt ? evt.pageY : ("clientY" in evt ? evt["clientY"] + de.scrollTop : null),

Could you integrate the other open PR to round pointer/mouse coordinates?

---
Refers to: src/plugins/pointer/mouse.ts:13 in e012cef. [](commit_id = e012cefb00776988b36cec9ba325c1dad8eaefc9, deletion_comment = False)

types/config.d.ts
```diff
@@ -26,15 +26,6 @@ export interface IConfig {
     // If set to 'true', clarity-js will NOT write CID (Clarity ID) cookie
     disableCookie?: boolean;
 
-    // If set to false, text on the page will be masked with asterisks
-    showText?: boolean;
-
-    // If set to false, links on the page will be masked with asterisks
-    showLinks?: boolean;
```
toby-walker: >showLinks [](start = 4, length = 9)

Is this always done now?

types/config.d.ts
```diff
@@ -26,15 +26,6 @@ export interface IConfig {
...
-
-    // If set to false, src of images won't be captured
-    showImages?: boolean;
```
toby-walker: >showImages [](start = 4, length = 10)

Is this always done now?

src/plugins/layout/nodeinfo.ts
ssivov: The logic guarantees it, so unless there is a bug, parent will always be set and processed before child is processed. That said, with our default masking 'on', our risk in case of a bug now becomes just accidentally masking something extra, which wouldn't hurt us or users. Traversing all the way up the DOM for every node is definitely more straight-forward, but is also a perf nightmare.

---
In reply to: [260854704](https://github.com/Microsoft/clarity-js/pull/137#discussion_r260854704) [](ancestors = 260854704)

types/config.d.ts
ssivov: Show links was a lousy option, which didn't really work out that great. There is no more special treatment for links. If webmaster wants to see his links, he'd have to set an attribute on them.

---
In reply to: [260855408](https://github.com/Microsoft/clarity-js/pull/137#discussion_r260855408) [](ancestors = 260855408)

types/config.d.ts
ssivov: images are not masked now

---
In reply to: [260855495](https://github.com/Microsoft/clarity-js/pull/137#discussion_r260855495) [](ancestors = 260855495)

ssivov:
>     y: "pageY" in evt ? evt.pageY : ("clientY" in evt ? evt["clientY"] + de.scrollTop : null),

Let's just have Sarvesh merge that one, no need to duplicate?

---
In reply to: [467950926](https://github.com/Microsoft/clarity-js/pull/137#issuecomment-467950926) [](ancestors = 467950926)

---
Refers to: src/plugins/pointer/mouse.ts:13 in e012cef. [](commit_id = e012cefb00776988b36cec9ba325c1dad8eaefc9, deletion_comment = False)

src/plugins/layout/nodeinfo.ts
ssivov: isCssText function? It's in states/text.ts

---
In reply to: [260854116](https://github.com/Microsoft/clarity-js/pull/137#discussion_r260854116) [](ancestors = 260854116)

PR #138, opened 2019-02-28
Allow capturing style sheet rules from all style nodes (through config)

ssivov:


src/plugins/layout/states/style.ts
```diff
@@ -31,3 +32,17 @@ export function getCssRules(element: HTMLStyleElement): string[] {
...
 }
+
+export function shouldCaptureCssRules(node: Node): boolean {
+    let captureCssRules = false;
+    if (node.nodeType === Node.ELEMENT_NODE && (node as Element).tagName === "STYLE") {
+        // If 'cssRules' is set to true, capture rules on all style nodes.
+        // Otherwise, optimistically hope that rules match the textContent,
+        // except for the case when there is no inner text. 'Empty' <style>
+        // node is a rather common way to add styles to the page by using
+        // <style>.insertRule API instead of appending css text children.
```
toby-walker: I couldn't quite follow this helpful comment. Can you clarify it? Might be faster to chat about quickly.

src/plugins/layout/states/ignore.ts
```diff
@@ -30,8 +30,7 @@ export function shouldIgnoreNode(node: Node, parentInfo: INodeInfo): boolean {
             // If we capture CSSRules on style elements, ignore its text children nodes
             // Since iterating over css rules can be 100X slower on certain browsers,
             // we limit ignoring text nodes to STYLE elements with empty text content
-            let parent = node.parentNode as Element;
-            if (parent && parent.tagName === "STYLE" && parent.textContent.length === 0) {
+            if (parentInfo && parentInfo.captureCssRules) {
                 ignore = true;
             }
```
toby-walker: Missing break at end of case.

src/plugins/layout/states/style.ts
ssivov: Well, that means i should definitely rephrase it :) But for your clarity

a. If config.cssRules == true, for any style node, capture it's sheet rules and ignore text child nodes
b. If config.cssRules == false, only "capture rules and ignore text children" for style nodes that have no children. If there is a style node with no children, it's very very likely that page does some direct css rule modifications through <style>.insertRule using this node, so if we don't capture it, the page will most likely be broken.

Does that make more sense?

---
In reply to: [261422191](https://github.com/Microsoft/clarity-js/pull/138#discussion_r261422191) [](ancestors = 261422191)

src/plugins/layout/states/ignore.ts
ssivov: Thanks

---
In reply to: [261422832](https://github.com/Microsoft/clarity-js/pull/138#discussion_r261422832) [](ancestors = 261422832)

src/plugins/layout/states/style.ts
ssivov: Updated comment

---
In reply to: [261423297](https://github.com/Microsoft/clarity-js/pull/138#discussion_r261423297) [](ancestors = 261423297,261422191)

src/plugins/layout/states/style.ts
```diff
@@ -31,3 +32,20 @@ export function getCssRules(element: HTMLStyleElement): string[] {
...
+    let captureCssRules = false;
+    if (node.nodeType === Node.ELEMENT_NODE && (node as Element).tagName === "STYLE") {
+        // Capturing CSS rules associated with a style node is more precise than relying on its childlist CSS text,
+        // because rules can be added/modified without affecting child text nodes, but it is also costly for performance.
+        // As a compromise, we are going to capture CSS rules in two cases:
+        // 1. If the config explicitly tells us to do so
+        // 2. If style node has no children. A rather common technique is to control page styles by inserting rules
+        // directly into the 'sheet holder' style node. So, when we encounter a style node with no inner text,
+        // chances are high that there are non-child-text rules associated with it and if we don't capture them,
+        // page visualization is likely to be broken. An popular example library that uses such technique is 'styled-components'.
```
toby-walker: >An [](start = 54, length = 2)

Nit: An -> A

PR #139, opened 2019-03-01
Adding discover/mutation performance instrumentation

ssivov:


src/plugins/layout/layout.ts
```diff
@@ -230,10 +246,28 @@ export default class Layout implements IPlugin {
...
       if (this.allowMutation()) {
-        this.processMutations(summary, time);
+        const stateGenStartTime = getTimestamp();
+        this.processMutations(summary, mutationStartTime);
+        stateGenCount = (
+          summary.newNodes.length
+          + summary.movedNodes.length
+          + summary.updatedNodes.length
+          + summary.removedNodes.length
+        );
```
ender336: do you think it would be useful at all to log each of these numbers rather than the aggregate? Just thinking about future debugging where it might be helpful to know at a glance we are getting pegged doing updates vs new nodes or something similar

types/instrumentation.d.ts
```diff
@@ -79,3 +80,19 @@ export interface IClarityActivateErrorState extends IInstrumentationEventState {
 export interface ITriggerState extends IInstrumentationEventState {
   key: string;
 }
+
+export interface IPerformanceState extends IInstrumentationEventState {
+  procedure: string;
+  duration: number;
```
ender336: >duration [](start = 2, length = 8)

might be nice to also include start time so we have a full picture

src/plugins/layout/layout.ts
ssivov: Hm, yeah, i think there is value in that. Right now we gen full state for any event (major perf hell, but nonetheless). However, such breakdown would give us an idea of how much perf we're wasting.

---
In reply to: [261791739](https://github.com/Microsoft/clarity-js/pull/139#discussion_r261791739) [](ancestors = 261791739)

types/instrumentation.d.ts
ssivov: I don't really see how that would be useful

---
In reply to: [261791885](https://github.com/Microsoft/clarity-js/pull/139#discussion_r261791885) [](ancestors = 261791885)

src/plugins/layout/layout.ts
ssivov: Done

---
In reply to: [261793020](https://github.com/Microsoft/clarity-js/pull/139#discussion_r261793020) [](ancestors = 261793020,261791739)

PR #145, opened 2019-07-30
Samart/payload truncation

ender336:


ender336:
will schema (delta encoding) break?

ender336:
need test case for analytics side to ensure it handles these cases gracefully

src/core.ts
```diff
@@ -119,12 +119,23 @@ export function bind(target: EventTarget, event: string, listener: EventListener
...
+        type: Instrumentation.OversizedEvent,
+        size: stateLength
+      }
+    }
+    : {
+      id: eventCount++,
+      time: isNumber(event.time) ? event.time : getTimestamp(),
+      type: event.type,
+      state: event.state
+    };
```
ssivov: This doesn't really solve the problem of the oversized payload, does it? Besides, this will post instrumentation even when inst is disabled by the config

ender336: Correct this is a mitigation, Clarity VNext that Sarvesh is working on is going to have the full solution. Intent here is to unblock as much as possible in the next month or two.

I can potentially change the event type away from Instrumentation or drop it all together when it is disabled by config, but this is an important notice to the server - so I don't think skipping it is really appropriate.

---
In reply to: [309257472](https://github.com/microsoft/clarity-js/pull/145#discussion_r309257472) [](ancestors = 309257472)

ender336:
no, event is changed before EventToArray is called

---
In reply to: [516584112](https://github.com/microsoft/clarity-js/pull/145#issuecomment-516584112) [](ancestors = 516584112)

src/core.ts
ssivov: Sure, just pointing out potential issues.

---
In reply to: [309351423](https://github.com/microsoft/clarity-js/pull/145#discussion_r309351423) [](ancestors = 309351423,309257472)

toby-walker: Please add a comment to the code that contains the gist of the discussion above. 

---
In reply to: [309414767](https://github.com/microsoft/clarity-js/pull/145#discussion_r309414767) [](ancestors = 309414767,309351423,309257472)

src/core.ts
```diff
@@ -119,12 +119,23 @@ export function bind(target: EventTarget, event: string, listener: EventListener
...
-  };
+  const stateLength = event.state ? JSON.stringify(event.state).length : -1;
+  let evtJson: IEvent = config.eventLimit && stateLength > config.eventLimit
+    ? {
+      id: eventCount++,
+      time: isNumber(event.time) ? event.time : getTimestamp(),
+      type: InstrumentationEventName,
+      state: {
+        type: Instrumentation.OversizedEvent,
+        size: stateLength
```
toby-walker: >stateLength [](start = 14, length = 11)

Include more information about the event. If you can send the event "shell" (the event types, etc.

PR #146, opened 2019-07-31
custom event support

ender336:


ender336:
test exercising this code on analytics side: [https://github.com/microsoft/clarity-analytics/pull/153](https://github.com/microsoft/clarity-analytics/pull/153)

ender336:
adding a few tests on this side as well

types/config.d.ts
```diff
@@ -67,4 +63,10 @@ export interface IConfig {
 
     // For pointer events, compute and record pointe event coordinates relative to its target's top left corner
     pointerTargetCoords?: boolean;
+
+    // Identifier of the user which generated this impression, if not provided Clarity will generate its own
```
toby-walker: >of [](start = 18, length = 2)

This seems weird to put inot the configuration.

types/config.d.ts
```diff
@@ -67,4 +63,10 @@ export interface IConfig {
...
+    userId?: string;
+
+    // Identifier for this impression, if not provided Clarity will generate its own
+    impressionId?: string;
```
toby-walker: >impressionId [](start = 4, length = 12)

Why is this in the configuration?

src/clarity.ts
```diff
@@ -19,3 +19,7 @@ export function stop(): void {
 export function trigger(key: string): void {
   onTrigger(key);
 }
+
+export function log(kvps: { [key: string]: any }): void {
```
ender336: >log [](start = 16, length = 3)

change to event

karma/setup/page.ts
```diff
@@ -41,3 +31,21 @@ export function cleanupPage(): void {
 
     fixture.cleanup();
 }
+
+export async function setupPageNoRunClarity(done: DoneFn): Promise<void> {
```
toby-walker: >setupPageNoRunClarity [](start = 22, length = 21)

What does this method name mean? I had a bit of trouble parsing the name.

src/clarity.ts
```diff
@@ -19,3 +19,11 @@ export function stop(): void {
 export function trigger(key: string): void {
   onTrigger(key);
 }
+
+export function event(kvps: { [key: string]: any }): void {
+  onCustomLog(kvps);
```
toby-walker: >onCustomLog [](start = 2, length = 11)

onCustomEvent

src/core.ts
```diff
@@ -188,6 +189,26 @@ export function onTrigger(key: string): void {
...
+      state: kvps,
+      time: getTimestamp(),
+    };
+    addEvent(event);
+  }
+}
+
+export function onSetPageInfo(pageId: string, userId: string): void {
+  // only allow setting pageId and userId if Clarity isn't currently running
+  if (state === State.Loaded || state === State.Unloaded) {
```
toby-walker: Should we trap the "else" case or at least log something? A general note would be to add diagnostic logging but have it disabled by default.

src/core.ts
```diff
@@ -188,6 +189,26 @@ export function onTrigger(key: string): void {
...
+    cid = userId;
+    setClarityCookie(userId);
+    impressionId = pageId;
```
toby-walker: >pageId [](start = 19, length = 6)

Both of these should set the value only if parameter is non-null/defined (to allow setting only one of them).

toby-walker:
Bump the version a major version as the change is not backwards compatible (it's an externally visible API change, breaks existing clients).

karma/setup/page.ts
ender336: hmm I'm trying to express that it sets up the page without running Clarity...let me try to rephrase and see it if makes it easier to follow

---
In reply to: [310726526](https://github.com/microsoft/clarity-js/pull/146#discussion_r310726526) [](ancestors = 310726526)

PR #147, opened 2019-07-31
ensuring remasking works

ender336:


src/plugins/layout/nodeinfo.ts
```diff
@@ -20,5 +21,11 @@ function shouldUnmaskNode(node: Node, parentInfo: INodeInfo): boolean {
       ? (node as Element).getAttribute(UnmaskAttribute) === "true"
       : false
   );
-  return hasUnmaskAttribute || parentUnmasked;
+  // mask attribute takes precedence over any unmask above
+  const hasMaskAttribute = (
```
toby-walker: I think you should change the logic here to inherit both the mask and unmask flags from the parent so that if you have a "mask" attribute (explicit) at a parent and there is no intervening unmask (and vice versa). Let's chat more if the comment isn't clear.

karma/tests/layout/privacy.ts
```diff
@@ -91,4 +92,32 @@ describe("Layout: Privacy Tests", () => {
         done();
     }));
 
+    it("checks that node value is masked event when explicitly masked if unmask attribute is applied to parent",
+        testAsync(async (done: DoneFn) => {
+            watch();
+            let valueString = "value";
+            let maskedValueString = mask(valueString);
+            let skipParent = document.createElement("div");
+            let parent = document.createElement("div");
```
toby-walker: Minor preference would be to just set the innerHTML and write this test as HTML fragment rather than a bunch of hard to read DOM calls.

src/plugins/layout/nodeinfo.ts
```diff
@@ -4,6 +4,7 @@ import { shouldCaptureCssRules } from "./states/style";
 import { isCssText } from "./states/text";
 
 export const UnmaskAttribute = "data-clarity-unmask";
```
toby-walker: >UnmaskAttribute [](start = 13, length = 15)

These should go into the publicly exported declaration for TypeScript. 
Also, we should have a read me or other documentation for the configuration and custom decoration attributes.

src/plugins/layout/nodeinfo.ts
ender336: making the change more clear

---
In reply to: [309714869](https://github.com/microsoft/clarity-js/pull/147#discussion_r309714869) [](ancestors = 309714869)

karma/tests/layout/privacy.ts
ender336: reasonable ask, but I'm going to punt refactoring our tests and let this go through as it's high pri

---
In reply to: [309715206](https://github.com/microsoft/clarity-js/pull/147#discussion_r309715206) [](ancestors = 309715206)

PR #148, opened 2019-09-20
Simplifying metrics and updating version

sarveshnagpal:
1. Addressing offline feedback.
2. Simplifying metrics code
3. Renaming stream -> analytics and backup -> playback
4. Upping the version and including it in core/version.ts


types/data.d.ts
```diff
@@ -32,7 +32,9 @@ export const enum Event {
     Performance = 25,
     ScriptError = 26,
     ImageError = 27,
-    LayoutSummary = 28
+    Layout = 28,
+    Crawl = 29,
+    Summary = 30
```
toby-walker: >Summary [](start = 4, length = 7)

Summary is too broad. What is it summarizing?

types/data.d.ts
```diff
@@ -41,13 +43,18 @@ export const enum Upload {
     Backup = 2
 }
 
+export const enum Flag {
```
toby-walker: >Flag [](start = 18, length = 4)

A bit too general. Maybe "BooleanFlag"?

types/data.d.ts
```diff
@@ -32,7 +32,9 @@ export const enum Event {
     Performance = 25,
     ScriptError = 26,
     ImageError = 27,
-    LayoutSummary = 28
+    Layout = 28,
+    Crawl = 29,
```
toby-walker: >Crawl [](start = 4, length = 5)

Crawl is the server action, not an event. Maybe RequiredResource?

types/data.d.ts
```diff
@@ -57,9 +64,9 @@ export interface IDecodedPayload {
     timestamp: number;
     ua: string;
     envelope: IEnvelope;
-    metrics: IDecodedMetric;
-    stream: IDecodedEvent[];
-    backup: IDecodedEvent[];
+    metrics: IMetric;
```
toby-walker: >IMetric [](start = 13, length = 7)

Should this be IMetric[]?

types/interaction.d.ts
```diff
@@ -2,6 +2,8 @@ export interface IPointer {
     target: number;
     x: number;
     y: number;
+    targetX?: number;
```
toby-walker: >targetX [](start = 4, length = 7)

What are target X/Y vs x/y?

src/metric/encode.ts
```diff
@@ -31,18 +31,9 @@ export default function(last: boolean = false): Token[] {
...
-        output.push(event.duration);
-    }
-
-    // Encode user specified marks
-    if (metrics.marks.length > 0) { output.push(MetricType.Marks); }
-    let marks = metrics.marks;
+    // Encode user specified tags
+    if (metrics.tags.length > 0) { output.push(MetricType.Tag); }
+    let marks = metrics.tags;
     for (let mark of marks) {
```
toby-walker: Remove marks? I think that's what we discussed on Friday.

decode/interaction.ts
```diff
@@ -17,6 +17,10 @@ export default function(tokens: Token[]): IDecodedEvent {
         case Event.TouchEnd:
         case Event.TouchMove:
             let pointerData: IPointer = { target: tokens[2] as number, x: tokens[3] as number, y: tokens[4] as number };
+            if (tokens.length > 6) {
```
toby-walker: When is this sent? Add comment.

decode/layout.ts
```diff
@@ -116,29 +130,38 @@ function process(node: any[] | number[], tagIndex: number): IDecodedNode {
...
-            selectorMap[id] = s;
-            return s;
+            if (s) { selectors.push({ id, checksum: hash(s), selector: s }); }
+            break;
+    }
+}
+
+function crawler(tag: string, attributes: IAttributes): void {
+    switch (tag) {
+        case "LINK":
```
toby-walker: >LINK [](start = 14, length = 4)

You should check the link is to a style sheet using the rel tag (rel=stylesheet) as there are many things that can be linked to.. Also record the media property? 
See: [https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link)

types/data.d.ts
sarveshnagpal: It is broad; it's summary for every event that was in the payload. It include the event, start and end. E.g. if were multiple mouse movement events between 323ms and 900ms then the summary event would be [9,232,900].

types/data.d.ts
sarveshnagpal: I will call it Resource.

types/data.d.ts
sarveshnagpal: It is currently an object instead of an array. We can track it either way but since we are sending only updated metrics back, we need to send back the index too. If you look at the interface - it's basically a dictionary:
interface IMetric {
    [key: number]: number;
}

types/interaction.d.ts
sarveshnagpal: Added a comment. It's relative (x,y) to the target element. We compute it for click events.

src/metric/encode.ts
sarveshnagpal: Yes gone per our discussion. Simplified tags by moving it as a top level event - it didn't have much purpose being within metrics.

decode/interaction.ts
sarveshnagpal: Done, added a comment in actual code file instead of the decode. We send it when it's a click, mainly because it's an expensive operation. Later when we wish to expand this - the decoding code will remain the same.

decode/layout.ts
sarveshnagpal: Good point, fixed.

PR #149, opened 2019-09-29
vNextBeta6: Type refactoring and addressing known issues

sarveshnagpal:
1. Strongly typing various event types to assist with analytics
2. Breaking payload into multiple streams - one per event type
3. Using discriminated unions to simplify types
4. Dropping "I" from interfaces as per guidelines by Typescript.
5. Stop sending redundant document information back to server
6. Simplifying metrics and folding them into Data and sending them as regular events
7. Dropping title from the Event.Page for security


decode/clarity.ts
```diff
@@ -1,40 +1,50 @@
 import version from "../src/core/version";
-import { Event, IAugmentation, IDecodedEvent, IDecodedPayload, IPayload, Token } from "../types/data";
-import data from "./data";
-import diagnostic from "./diagnostic";
-import envelope from "./envelope";
-import interaction from "./interaction";
+import { Augmentation, Event, Payload, Token } from "../types/data";
+import * as Decode from "../types/decode";
```
toby-walker: Style: I generally prefer to avoid "*" for important and name everything used explicitly as you did a few places here. It's easier then to see what comes from what. If you need to import a lot of stuff, something is off with the module boundaries.

decode/clarity.ts
```diff
@@ -1,40 +1,50 @@
...
-    let timestamp = augmentations ? augmentations.timestamp : Date.now();
-    let ua = augmentations ? augmentations.ua : (navigator && "userAgent" in navigator ? navigator.userAgent : "");
-    let payload: IDecodedPayload = { timestamp, ua, envelope: envelope(json.e), metrics: metric(json.m), analytics: [], playback: [] };
+export function decode(input: string | Payload, augmentations: Augmentation = null): Decode.DecodedPayload {
+    let json: Payload = typeof input === "string" ? JSON.parse(input) : input;
+    let envelope = data.envelope(json.e);
+    let timestamp = augmentations && augmentations.timestamp ? augmentations.timestamp : Date.now();
+    let ua = augmentations && augmentations.ua ? augmentations.ua : (navigator && "userAgent" in navigator ? navigator.userAgent : "");
+    let payload: Decode.DecodedPayload = { timestamp, envelope };
+    if (envelope.sequence === 1) { payload.ua = ua; }
```
toby-walker: This is a good change. We don't need to send the UA on each payload. Might consider just making this part of the page event as it isn't really payload property (logically).

decode/clarity.ts
```diff
@@ -1,40 +1,50 @@
...
+        data.summarize(entry);
         switch (entry[1]) {
-            case Event.Scroll:
-            case Event.Document:
-            case Event.Resize:
-            case Event.Selection:
-            case Event.Change:
+            case Event.Page:
+                if (payload.page === undefined) { payload.page = []; }
+                payload.page.push(data.decode(entry) as Decode.PageEvent);
```
toby-walker: >page [](start = 24, length = 4)

Can we change these so that we have a single pointer event and then whether touch or not touch is just a property. Makes handling these a bit clearer as the touch vs. non-touch isn't that central to handling them. So the y would be Pointer{Up,Down,Move,Cancel} for the corresponding events.

toby-walker:
>             case Event.RightClick:

Would prefer to just have a single Click event with properties indicating what kind it is. 

---
Refers to: decode/clarity.ts:54 in d55bcaa. [](commit_id = d55bcaaccf1ab80099e90366fd3c124e75910820, deletion_comment = False)

toby-walker:
Cod you please tag the commit with the version after you merge it so that's easier to start tracking this and go from version to commit?

decode/clarity.ts
```diff
@@ -46,83 +56,109 @@ export function decode(input: string | IPayload, augmentations: IAugmentation =
...
+                break;
+            case Event.Resize:
+                if (payload.resize === undefined) { payload.resize = []; }
+                payload.resize.push(interaction.decode(entry) as Decode.ResizeEvent);
+                break;
+            case Event.Selection:
+                if (payload.selection === undefined) { payload.selection = []; }
+                payload.selection.push(interaction.decode(entry) as Decode.SelectionEvent);
+                break;
+            case Event.Change:
```
toby-walker: Very vague event name. Can you rename to something clearer (even if this is what the DOM event is called). For example, InputChanged?

decode/clarity.ts
```diff
@@ -46,83 +56,109 @@ export function decode(input: string | IPayload, augmentations: IAugmentation =
...
+                payload.boxmodel.push(layout.decode(entry) as Decode.BoxModelEvent);
                 break;
             case Event.Discover:
             case Event.Mutation:
-                event = layout.decode(entry);
-                payload.playback.push(event);
+                if (payload.dom === undefined) { payload.dom = []; }
+                payload.dom.push(layout.decode(entry) as Decode.DomEvent);
                 break;
             case Event.Checksum:
```
toby-walker: >Checksum [](start = 23, length = 8)

CAn we rename this to LayoutSummary or something more indicative? "Hash" would be better than "Checksum" (which usually is a check for errors in the data).

decode/data.ts
```diff
@@ -1,23 +1,76 @@
...
         case Event.Page:
-            let page: IPageData = {
+            let page: PageData = {
                 timestamp: tokens[2] as number,
                 elapsed: tokens[3] as number,
                 url: tokens[4] as string,
-                title: tokens[5] as string,
-                referrer: tokens[6] as string
+                referrer: tokens[5] as string,
+                lean: tokens[6] as number,
```
toby-walker: >lean [](start = 16, length = 4)

Is this a boolean? If so can you type it as something that indicates that is the case -- an enum with values for True and False (or On/Off or whatever you think makes sense).

decode/data.ts
```diff
@@ -1,23 +1,76 @@
...
+}
+
+export function decode(tokens: Token[]): DataEvent {
     let time = tokens[0] as number;
     let event = tokens[1] as Event;
     switch (event) {
         case Event.Page:
-            let page: IPageData = {
+            let page: PageData = {
                 timestamp: tokens[2] as number,
```
toby-walker: >timestamp [](start = 16, length = 9)

What are these fields (timestamp and elapsed)?

decode/data.ts
```diff
@@ -1,23 +1,76 @@
...
+    return {
+        elapsed: tokens[0] as number,
+        sequence: tokens[1] as number,
+        version: tokens[2] as string,
+        projectId: tokens[3] as string,
+        userId: tokens[4] as string,
+        sessionId: tokens[5] as string,
+        pageId: tokens[6] as string,
+        upload: tokens[7] as Upload,
+        end: tokens[8] as State
```
toby-walker: >State [](start = 26, length = 5)

State is very broad name. Could you indicate what kind of state this represents and prefix the name with that?

decode/data.ts
```diff
@@ -1,23 +1,76 @@
...
+    };
+}
+
+export function summarize(entry: Token[]): void {
```
toby-walker: I see this in the data but wasn't sure why we send it. What use did you have in mind?

toby-walker:
>                 resources.push({ tag, url: attributes["href"]});

Can you add the relationship (rel attribute) to this event? For future use. 

---
Refers to: decode/layout.ts:168 in d55bcaa. [](commit_id = d55bcaaccf1ab80099e90366fd3c124e75910820, deletion_comment = False)

src/core/version.ts
```diff
@@ -1,2 +1,2 @@
-let version = "1.0.0-b5";
+let version = "1.0.0-b6";
```
toby-walker: Very minor; Can you read this from the package rather than repeating it?

toby-walker:
>             xhr.send(data);

Important: Add retry. Send gif or something when error happens. 

---
Refers to: src/data/upload.ts:74 in d55bcaa. [](commit_id = d55bcaaccf1ab80099e90366fd3c124e75910820, deletion_comment = False)

decode/layout.ts
```diff
@@ -1,58 +1,60 @@
...
 }
 
-export function decode(tokens: Token[]): IDecodedEvent {
+export function decode(tokens: Token[]): LayoutEvent {
     let time = lastTime = tokens[0] as number;
     let event = tokens[1] as Event;
-    let decoded: IDecodedEvent = {time, event, data: []};
 
     switch (event) {
         case Event.Document:
```
toby-walker: >Document [](start = 19, length = 8)

I saw lot of examples of this event with the same values repeated for the same page (i.e. identical width/height). When do we send it?

decode/clarity.ts
sarveshnagpal: Good suggestion. Resolved and broke decode types into sub-types.

decode/clarity.ts
sarveshnagpal: Moved it to Event.Page instead of augmentations. Also, removed the concept of augmentations for now.

decode/clarity.ts
sarveshnagpal: There's not a lot of overlap between mouse & touch events:
- MouseUp | MouseDown | MouseMove | MouseWheel
- TouchStart | TouchCancel | TouchMove | TouchEnd

So the only thing that is common is *Move. Let's discuss this offline, won't fix for now.

decode/clarity.ts
sarveshnagpal: Done => InputChange

decode/clarity.ts
sarveshnagpal: Renamed it to Hash

decode/data.ts
sarveshnagpal: Yes good point - using enum now.

decode/data.ts
sarveshnagpal: timestamp is absolute unix time. The idea is that if we needed to, we could turn relative "time" values in payload to absolute ones. I've removed elapsed since it was adding confusion and I already have "time" in the event wrapper.

decode/data.ts
sarveshnagpal: Let me take this offline. If we don't want it I can remove it.

decode/data.ts
sarveshnagpal: Thanks and called it BooleanFlag.

src/core/version.ts
sarveshnagpal: I'll get back to this. Don't want to introduce side effects by importing package.json without testing this out thoroughly.

decode/layout.ts
sarveshnagpal: Already fixed the repetitions. No longer an issue.

src/interaction/encode.ts
```diff
@@ -85,7 +84,7 @@ export default function(type: Event): void {
             }
             scroll.reset();
             break;
-        case Event.Visibility:
+        case Event.Visible:
```
toby-walker: Preferred the old name. Minor point.

src/data/upload.ts
```diff
@@ -48,33 +55,51 @@ export function end(): void {
...
-    return `{"e":${payload.e},"m":${payload.m},"d":${payload.d}}`;
+function stringify(payload: EncodedPayload): string {
+    return `{"e":${payload.e},"d":${payload.d}}`;
 }
 
-function send(data: string, last: boolean = false): void {
+function send(data: string, sequence: number = null, last: boolean = false): void {
     if (config.url.length > 0) {
         if (last && "sendBeacon" in navigator) {
             navigator.sendBeacon(config.url, data);
```
toby-walker: >navigator [](start = 12, length = 9)

Future: Should fall back to XHR if this causes an exception -- e.g.., if the payload is too large for a beacon.

decode/clarity.ts
toby-walker: Actually there is as shown by the pointer event enent model where MouseUp <-> TouchEnd, MouseDown <-> TouchStart MouseMove<->TouchMove, MouseCacnel/PointerCancle<->TouchCancel PointerError/
Agree, we can refine this offline for next iteration. 

---
In reply to: [329382328](https://github.com/microsoft/clarity-js/pull/149#discussion_r329382328) [](ancestors = 329382328)

src/interaction/encode.ts
sarveshnagpal: Updated the name, thanks.

src/data/upload.ts
sarveshnagpal: This is a good point and will think over it before addressing it in the next version.

PR #151, opened 2019-10-07
Bug fixes and data quality improvements over vBeta7

sarveshnagpal:
- Decoding fix to look up parent selectors
- Script errors
- Honoring upload handler even in case of recovery mode

### DIFF
--- a/decode/data.ts
+++ b/decode/data.ts
@@ -32,7 +32,7 @@ export function decode(tokens: Token[]): DataEvent {
             let upload: UploadData = { sequence: tokens[2] as number, attempts: tokens[3] as number, status: tokens[4] as number};
             return { time, event, data: upload };
         case Event.Metric:
-            let i = 2;
+            let i = 2; // Start from 3rd index since first two are used for time & event
             let metrics: MetricData = {};
             while (i < tokens.length) {
                 metrics[tokens[i++] as number] = tokens[i++] as number;
--- a/decode/data.ts
+++ b/decode/data.ts
@@ -32,7 +32,7 @@ export function decode(tokens: Token[]): DataEvent {
             let upload: UploadData = { sequence: tokens[2] as number, attempts: tokens[3] as number, status: tokens[4] as number};
             return { time, event, data: upload };
         case Event.Metric:
-            let i = 0;
+            let i = 2;
             let metrics: MetricData = {};
             while (i < tokens.length) {
                 metrics[tokens[i++] as number] = tokens[i++] as number;
--- a/decode/layout.ts
+++ b/decode/layout.ts
@@ -5,6 +5,8 @@ import { DomData, LayoutEvent } from "../types/decode/layout";
 import { Attributes, BoxModelData, DocumentData, HashData, ResourceData } from "../types/layout";
 
 const ID_ATTRIBUTE = "data-clarity";
+const SVG_PREFIX: string = "svg:";
+
 let placeholderImage = "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR4nGNiOAMAANUAz5n+TlUAAAAASUVORK5CYII=";
 export let hashes: { [key: number]: HashData } = {};
 export let resources: ResourceData[];
@@ -157,9 +159,10 @@ function getHash(id: number, tag: string, path: string, attributes: Attributes):
         case "*D":
             break;
         default:
+            tag = tag.indexOf(SVG_PREFIX) === 0 ? tag.substr(SVG_PREFIX.length) : tag;
             let s = path && path.length > 0 ? path + tag : tag;
             if ("id" in attributes && attributes["id"].length > 0) { s = `${tag}#${attributes["id"]}`; }
-            if ("class" in attributes && attributes["class"].length > 0) { s += `.${attributes["class"].trim().split(" ").join(".")}`; }
+            if ("class" in attributes && attributes["class"].length > 0) { s += `.${attributes["class"].trim().split(/\s+/).join(".")}`; }
             if (ID_ATTRIBUTE in attributes) { s = `*${attributes[ID_ATTRIBUTE]}`; }
             if (s) { hashes[id] = { id, hash: generateHash(s), selector: s }; }
             break;
--- a/decode/layout.ts
+++ b/decode/layout.ts
@@ -1,8 +1,9 @@
 import generateHash from "../src/data/hash";
 import { resolve } from "../src/data/token";
+import selector from "../src/layout/selector";
 import { Event, Token } from "../types/data";
 import { DomData, LayoutEvent } from "../types/decode/layout";
-import { Attributes, BoxModelData, Constant, DocumentData, HashData, ResourceData } from "../types/layout";
+import { Attributes, BoxModelData, DocumentData, HashData, ResourceData } from "../types/layout";
 
 let placeholderImage = "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR4nGNiOAMAANUAz5n+TlUAAAAASUVORK5CYII=";
 export let hashes: { [key: number]: HashData } = {};
@@ -108,7 +109,7 @@ function process(node: any[] | number[], tagIndex: number): DomData {
     let hasAttribute = false;
     let attributes = {};
     let value = null;
-    let path = output.parent in hashes ? `${hashes[output.parent].selector}>` : null;
+    let prefix = output.parent in hashes ? `${hashes[output.parent].selector}>` : null;
 
     for (let i = tagIndex + 1; i < node.length; i++) {
         let token = node[i] as string;
@@ -117,7 +118,7 @@ function process(node: any[] | number[], tagIndex: number): DomData {
         if (i === (node.length - 1) && output.tag === "STYLE") {
             value = token;
         } else if (lastChar === ">" && keyIndex === -1) {
-            path = token;
+            prefix = token;
         } else if (output.tag !== "*T" && keyIndex > 0) {
             hasAttribute = true;
             let k = token.substr(0, keyIndex);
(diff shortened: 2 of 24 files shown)


decode/layout.ts
```diff
@@ -107,7 +109,7 @@ function process(node: any[] | number[], tagIndex: number): DomData {
     let hasAttribute = false;
     let attributes = {};
     let value = null;
-    let path = output.parent in hashes ? `${hashes[output.parent]}>` : null;
+    let path = output.parent in hashLookup ? `${hashes[hashLookup[output.parent]].selector}>` : null;
```
toby-walker: Why is the lookup/cache needed? Wouldn't the same selector or index always return the same value? Can you add a comment that gives the answer to this?

decode/layout.ts
```diff
@@ -156,7 +158,10 @@ function getHash(id: number, tag: string, path: string, attributes: Attributes):
             if ("id" in attributes) { s = `${tag}#${attributes["id"]}`; }
             if ("class" in attributes) { s += `.${attributes["class"].trim().split(" ").join(".")}`; }
             if (ID_ATTRIBUTE in attributes) { s = `*${attributes[ID_ATTRIBUTE]}`; }
-            if (s) { hashes.push({ id, hash: generateHash(s), selector: s }); }
+            if (s) {
+                hashLookup[id] = hashes.length;
```
toby-walker: Why do you need two tables (hashlookup and hashes)? Could you not combine into a single table?

toby-walker:
>             encode(Event.Upload);

Suggestion: This doesn't record the event or details of failures. Can you encode information about the failed request status codes? I had this feedback in the list of things based on reviewing data for this iteration. 

---
Refers to: src/data/upload.ts:96 in 6aa7245. [](commit_id = 6aa72450e86a0e0ec34a529540a22170a862cc4c, deletion_comment = False)

decode/data.ts
```diff
@@ -32,7 +32,7 @@ export function decode(tokens: Token[]): DataEvent {
             let upload: UploadData = { sequence: tokens[2] as number, attempts: tokens[3] as number, status: tokens[4] as number};
             return { time, event, data: upload };
         case Event.Metric:
-            let i = 0;
+            let i = 2;
```
toby-walker: Would you mind adding a bit ore to the description describing the issues this update resolves?

src/interaction/selection.ts
```diff
@@ -18,15 +18,24 @@ export function start(): void {
 function recompute(): void {
     let s = document.getSelection();
 
-    if (selection !== null && data.start !== null && data.start !== getId(s.anchorNode)) {
+    // Bail out if we don't have a valid selection
```
toby-walker: Is this what happens when a user clears their selection? Should we send an update to indicate that?

src/interaction/selection.ts
```diff
@@ -18,15 +18,24 @@ export function start(): void {
...
+    if (s === null) { return; }
+
+    let anchorNode = getId(s.anchorNode);
+    let focusNode = getId(s.focusNode);
+
+    // Bail out if we got valid selection but not valid nodes
+    if (anchorNode === null && focusNode === null) { return; }
```
toby-walker: >null [](start = 23, length = 4)

How could that happen?

src/interaction/selection.ts
```diff
@@ -18,15 +18,24 @@ export function start(): void {
...
+
+    if (selection !== null && data.start !== null && data.start !== anchorNode) {
```
toby-walker: Minor but you don't need the selection !== null here as you checked for it above.

decode/layout.ts
sarveshnagpal: I simplified this and hashLookup no longer exists.

decode/data.ts
sarveshnagpal: Yes sure; added a comment.

src/interaction/selection.ts
sarveshnagpal: Technically what I've checked above is current selection (s), while this variable refers to last stored selection (selection). Since last is computed of current at some point - we don't need this additional check. I've also renamed the variables to make it clear.

src/interaction/selection.ts
sarveshnagpal: Since selection is not sent as part of an event property, like other events, and we have to poll for the selection after the fact - there's always a chance that we are out of date. We are just ensuring that we are not dealing with a stale selection here.

src/interaction/selection.ts
sarveshnagpal: Added comment - basically, selection change event fires in Edge even on right click or click. If there was no previous selection, document.getSelection() returns null nodes.

src/layout/node.ts
```diff
@@ -3,6 +3,8 @@ import config from "@src/core/config";
 import * as dom from "./dom";
 
 const IGNORE_ATTRIBUTES = ["title", "alt", "onload", "onfocus"];
+const SVG_NAMESPACE: string = "http://www.w3.org/2000/svg";
+const SVG_PREFIX: string = "svg:";
```
toby-walker: You repeat this in the dom.ts file. Remove one of them.

src/layout/dom.ts
```diff
@@ -199,14 +201,20 @@ function selector(id: number, data: NodeInfo, parent: string): string {
...
             let value = getValue(id);
             let ex = value ? value.selector : null;
             let attributes = "attributes" in data ? data.attributes : {};
-            let current = "id" in attributes ? `${data.tag}#${attributes.id}` : `${parent}>${data.tag}`;
-            if ("class" in attributes) { current = `${current}.${attributes.class.trim().split(" ").join(".")}`; }
-            if (ID_ATTRIBUTE in attributes) { current = `*${attributes[ID_ATTRIBUTE]}`; }
-            if (current !== ex && selectorMap.indexOf(id) === -1) { selectorMap.push(id); }
-            return current;
+            let tag = data.tag.indexOf(SVG_PREFIX) === 0 ? data.tag.substr(SVG_PREFIX.length) : data.tag;
+            let s = "id" in attributes && attributes["id"].length > 0 ? `${tag}#${attributes.id}` : `${parentSelector}>${tag}`;
```
toby-walker: No change but worth discussing this logic given how important it is for analytics.

src/layout/boxmodel.ts
```diff
@@ -50,9 +50,11 @@ export function updates(): BoxModelData[] {
 
 export function relative(x: number, y: number, element: Element): number[] {
     if (x && x >= 0 && y && y >= 0 && element) {
-        let box = layout(element);
+        let scrollX = "pageXOffset" in window ? window.pageXOffset : document.documentElement.scrollLeft;
+        let scrollY = "pageYOffset" in window ? window.pageYOffset : document.documentElement.scrollTop;
+        let box = layout(element, scrollX, scrollY);
```
toby-walker: Use de-structuring assignment to write

let [x, y, width, height] = layout(....)

Rather than box[0] as here.

src/interaction/selection.ts
```diff
@@ -16,27 +16,38 @@ export function start(): void {
 }
 
 function recompute(): void {
-    let s = document.getSelection();
+    let current = document.getSelection();
 
-    if (selection !== null && data.start !== null && data.start !== getId(s.anchorNode)) {
+    // Bail out if we don't have a valid selection
+    if (current === null) { return; }
```
toby-walker: If current === null and previous != null then should we send an event?

src/interaction/pointer.ts
```diff
@@ -12,16 +12,16 @@ let timeout: number = null;
...
-    bind(document, "mouseup", mouse.bind(this, Event.MouseUp));
-    bind(document, "mousemove", mouse.bind(this, Event.MouseMove));
-    bind(document, "mousewheel", mouse.bind(this, Event.MouseWheel));
-    bind(document, "dblclick", mouse.bind(this, Event.DoubleClick));
-    bind(document, "click", mouse.bind(this, Event.Click));
-    bind(document, "touchstart", touch.bind(this, Event.TouchStart));
-    bind(document, "touchend", touch.bind(this, Event.TouchEnd));
-    bind(document, "touchmove", touch.bind(this, Event.TouchMove));
-    bind(document, "touchcancel", touch.bind(this, Event.TouchCancel));
+    bind(document, "mousedown", mouse.bind(this, Event.MouseDown), true);
```
toby-walker: What's the true mean? Add comment.

decode/render.ts
```diff
@@ -3,10 +3,12 @@ import { DomData } from "../types/decode/layout";
 import { InputChangeData, PointerData, ResizeData, ScrollData, SelectionData } from "../types/interaction";
 import { BoxModelData } from "../types/layout";
 
+const SVG_NAMESPACE: string = "http://www.w3.org/2000/svg";
```
toby-walker: Don't deduplciate this here. Define it one place and reference. It's in at least 3 files of the code. 
I'm very strong about "DRY" (don't repeat yourself). Hence harping on this.

decode/layout.ts
```diff
@@ -134,7 +138,10 @@ function process(node: any[] | number[], tagIndex: number): DomData {
         }
     }
 
-    getHash(output.id, output.tag, path, attributes);
+    // Do not compute hash for nodes that are disconnected from DOM (i.e. no parent)
+    // The only exception is HTML element, which is the root element of DOM tree
+    if (output.parent !== null || output.tag === "HTML") { getHash(output.id, output.tag, path, attributes); }
```
toby-walker: Nit; Move to separate line.

decode/layout.ts
```diff
@@ -152,11 +159,12 @@ function getHash(id: number, tag: string, path: string, attributes: Attributes):
         case "*D":
             break;
         default:
+            tag = tag.indexOf(SVG_PREFIX) === 0 ? tag.substr(SVG_PREFIX.length) : tag;
             let s = path && path.length > 0 ? path + tag : tag;
```
toby-walker: Isn't this repeating logic for computing the selector? Can you move it one location?

src/interaction/selection.ts
sarveshnagpal: We will automatically send the previous based on the timeout that we scheduled earlier. So previous will go out automatically if there's no valid current within the look ahead period.

src/layout/dom.ts
sarveshnagpal: This is all based on Swaathee's feedback as she is prototyping Heatmaps.

sarveshnagpal: And, hoping to discuss this more as part of the design review later.

src/interaction/pointer.ts
sarveshnagpal: It's defined in Typescript schema (capture the event).

src/layout/node.ts
sarveshnagpal: Thanks, done. Changed all const strings into const enum and now we don't have to waste bytes sharing the variable name across files - which don't get minified later.

decode/render.ts
sarveshnagpal: Done. Refactored into const enum.

decode/layout.ts
sarveshnagpal: Done

src/layout/boxmodel.ts
sarveshnagpal: Done, looks much cleaner. Thanks.

src/interaction/selection.ts
toby-walker: Wasns't obvious. Can you add more comment about the logic? It's a bit opaque. now. 

---
In reply to: [334240742](https://github.com/microsoft/clarity-js/pull/151#discussion_r334240742) [](ancestors = 334240742)

src/interaction/pointer.ts
toby-walker: Maybe but it's surprisingly hard to find documentation about this -- e.g., the bind page on MDN doesn't make it clear. What does "capture the event" mean? Is this "addEventListener" (which has clearer documentation and seems to be what you are doing here).  
[https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener)

It would be clearer if you worte this as:

document.addEventListender((, <event, mouse.bind(....), {capture: true})

Boiolean optoins are too opauqe and should be avoided. 

BTW, should this not also add teh "passive:true" optimization to help with scrolling, etc. 


---
In reply to: [334240802](https://github.com/microsoft/clarity-js/pull/151#discussion_r334240802) [](ancestors = 334240802)

decode/layout.ts
toby-walker: Echoing this comment again. Similarly looking logic in two places. Refactor. 

---
In reply to: [334058330](https://github.com/microsoft/clarity-js/pull/151#discussion_r334058330) [](ancestors = 334058330)

sarveshnagpal: Done. I left the comment open because I wanted to get back to it and was on my list as well. It was getting redundant to write same things at two different places and keep them in sync. We can now expand on selector logic without worry about encoding / decoding.

PR #153, opened 2019-10-17
Incremental bug fixes over v1.0.0-beta8

sarveshnagpal:
- Fixing script errors in decode function
- Changing summary threshold to 30ms from 250ms
- Removing local storage backup option to fix client script errors
- Null check around observer.disconnect() call

### DIFF
--- a/decode/render.ts
+++ b/decode/render.ts
@@ -129,7 +129,7 @@ export function markup(data: DomData[], iframe: HTMLIFrameElement): void {
             case "*T":
                 let textElement = element(node.id);
                 textElement = textElement ? textElement : doc.createTextNode(null);
-                textElement.nodeValue = node.value;
+                textElement.nodeValue = unmask(node.value);
                 insert(node, parent, textElement, next);
                 break;
             case "HTML":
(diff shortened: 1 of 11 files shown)


decode/render.ts
```diff
@@ -218,10 +218,11 @@ function setAttributes(node: HTMLElement, attributes: object): void {
     for (let attribute in attributes) {
         if (attributes[attribute] !== undefined) {
             try {
+                let v = unmask(attributes[attribute]);
                 if (attribute.indexOf("xlink:") === 0) {
-                    node.setAttributeNS("http://www.w3.org/1999/xlink", attribute, attributes[attribute]);
+                    node.setAttributeNS("http://www.w3.org/1999/xlink", attribute, v);
                 } else {
-                    node.setAttribute(attribute, attributes[attribute]);
+                    node.setAttribute(attribute, v);
```
toby-walker: >attribute [](start = 38, length = 9)

Will masking attribute values affect playback for some attributes?

sarveshnagpal: No, shouldn't have an impact of the playback. Before rendering we are attempting to unmask the text if it may already have been masked during the instrumentation.

PR #154, opened 2019-10-19
Bug fixes and feature additions in v1.0.0-beta.10

sarveshnagpal:
- Enriching selectors with nth-of-type position. E.g. OL#b_results>LI:nth-of-type(8).b_algo>H2

To read more about nth-of-type, please see: https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-of-type

### DIFF
--- a/src/layout/dom.ts
+++ b/src/layout/dom.ts
@@ -55,6 +55,7 @@ export function add(node: Node, data: NodeInfo, source: Source): void {
         parent: parentId,
         next: nextId,
         children: [],
+        position: null,
         data,
         selector: "",
         metadata: { active: true, boxmodel: false, masked }
@@ -118,10 +119,27 @@ export function update(node: Node, data: NodeInfo, source: Source): void {
     }
 }
 
+function position(parent: NodeValue, child: NodeValue): number {
+    let tag = child.data.tag;
+    // Find relative position of the element to generate :nth-of-type selector
+    // We restrict relative positioning to handful of tags for now.
+    if (parent && (tag === "DIV" || tag === "TR" || tag === "P" || tag === "LI")) {
+        child.position = 1;
+        let idx = parent ? parent.children.indexOf(child.id) : -1;
+        while (idx-- > 0) {
+            let sibling = values[parent.children[idx]];
+            if (child.data.tag === sibling.data.tag) { child.position = sibling.position + 1; }
+            break;
+        }
+    }
+    return child.position;
+}
+
 function updateSelector(value: NodeValue): void {
-    let prefix = value.parent && value.parent in values ? `${values[value.parent].selector}>` : null;
+    let parent = value.parent && value.parent in values ? values[value.parent] : null;
+    let prefix = parent ? `${parent.selector}>` : null;
     let ex = value.selector;
-    let current = selector(value.data.tag, prefix, value.data.attributes);
+    let current = selector(value.data.tag, prefix, value.data.attributes, position(parent, value));
     if (current !== ex && selectorMap.indexOf(value.id) === -1) { selectorMap.push(value.id); }
     value.selector = current;
 }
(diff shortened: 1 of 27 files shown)


types/decode/layout.d.ts
```diff
@@ -16,6 +16,7 @@ export interface DomData {
     parent: number;
     next: number;
     tag: string;
+    position: number;
```
toby-walker: >number [](start = 14, length = 6)

Very minor but could we declare a type alias to distinguish integer number versus floating point? It makes it easier to know for a human reader. It will have no impacting on typing though.

src/layout/selector.ts
```diff
@@ -15,7 +16,7 @@ export default function(tag: string, prefix: string, attributes: Attributes): st
         default:
             if (prefix === null) { return empty; }
             tag = tag.indexOf(Constant.SVG_PREFIX) === 0 ? tag.substr(Constant.SVG_PREFIX.length) : tag;
-            let s = "id" in attributes && attributes["id"].length > 0 ? `${tag}#${attributes.id}` : `${prefix}${tag}`;
+            let s = "id" in attributes && attributes["id"].length > 0 ? `${tag}#${attributes.id}` : `${prefix}${tag}${suffix}`;
```
toby-walker: >${suffix} [](start = 116, length = 9)

Add a delimiter here between the tag and suffix.

src/layout/encode.ts
```diff
@@ -61,7 +61,7 @@ export default async function(type: Event): Promise<void> {
                                 tokens.push(value.id);
                                 if (value.parent && active) { tokens.push(value.parent); }
                                 if (value.next && active) { tokens.push(value.next); }
-                                metadata.push(data[key]);
+                                metadata.push(value.position ? `${data[key]}.${value.position}` : data[key]);
```
toby-walker: >. [](start = 76, length = 1)

Seems too class to notation used for class name in CSS selectors. Could we use another notation or separator like @ to similar.

src/layout/dom.ts
```diff
@@ -118,10 +119,27 @@ export function update(node: Node, data: NodeInfo, source: Source): void {
     }
 }
 
+function position(parent: NodeValue, child: NodeValue): number {
+    let tag = child.data.tag;
+    // Find relative position of the element to generate :nth-of-type selector
+    // We restrict relative positioning to handful of tags for now.
+    if (parent && (tag === "DIV" || tag === "TR" || tag === "P" || tag === "LI")) {
```
toby-walker: >TR [](start = 45, length = 2)

What about TD?

types/decode/layout.d.ts
sarveshnagpal: I could but in clarity anytime we use a number it's always an integer. We don't capture floats unless you believe I am missing something here.

toby-walker:
Could you update the code review to include the new version (i.e .b10) so it's easier to see what's in the PR. More useful to know that than the older version.

src/layout/selector.ts
sarveshnagpal: Similar to prefix, suffix contains the delimiter. If suffix is empty, we don't want a delimiter. This is to ensure that we end up with valid selectors in the end.

src/layout/encode.ts
sarveshnagpal: Yes good idea - let me change it.

sarveshnagpal: I'll use ~ because we don't have to encode it in the URL; if we choose to later.

src/layout/dom.ts
sarveshnagpal: Swaathee did some initial analysis to restrict the tags to select few to strike a balance between being unique enough and being too specific. She will cover this in her doc. I am assuming we will go through iterations here but going with the list she shared with me based on her research so far. So won't fix for now but definitely open to making changes after Swaathee's design review.

PR #156, opened 2019-11-04
wapring decode-js to expose the module

linrosales:
decode-js is exported to be used in bundle

### DIFF
--- a/webpack/configs/prod.ts
+++ b/webpack/configs/prod.ts
@@ -12,7 +12,7 @@ const ProdConfig: webpack.Configuration = {
 
     entry: {
         clarity: "./webpack/globalize.ts",
-        decode: "./decode/clarity.ts"
+        decode: "./webpack/globalizeDecoder.ts"
     },
 
     output: {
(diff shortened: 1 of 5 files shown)


webpack/configs/prod.ts
```diff
@@ -12,7 +12,7 @@ const ProdConfig: webpack.Configuration = {
 
     entry: {
         clarity: "./webpack/globalize.ts",
-        decode: "./decode/clarity.ts"
+        decode: "./webpack/globalize.ts",
```
sarveshnagpal: We do need two different files like you had before. We shouldn't have any reference to "decode" code in our primary clarity bundle that will get included for instrumentation. Decode is optional and should use a separate global file.

webpack/globalizeDecode.ts
```diff
@@ -0,0 +1,12 @@
+import * as decode from "@decode/clarity";
+import * as clarity from "@src/clarity";
+
+// When built with webpack for prod, compiled clarity-js bundle doesn't expose the module anywhere on the page.
+// Since we need clarity-js having the decode functionality to be available globally, we can create a wrapper
+// module that would assign clarity to window.
+if (typeof window !== "undefined") {
+    if ((window as any).clarity === undefined) {
+        (window as any).clarity = clarity;
```
sarveshnagpal: There won’t be a valid clarity variable. So the follow line would throw an error since we will attempt to assign a property “decode” of a null object. Can you please change this to { decode: decode } and the line below should be in else clause.

PR #157, opened 2019-11-05
Bug fixes and data quality improvements in 1.0.0-beta.11

sarveshnagpal:
- Updating hash to 9 chars based on v10 conflicts
- Resolving the issue with missing targets 
- Simplify lean mode to remove Event.Hash

More fixes in progress and will keep updating the PR.

### DIFF
--- a/decode/clarity.ts
+++ b/decode/clarity.ts
@@ -1,11 +1,11 @@
 import version from "../src/core/version";
 import { Event, Payload, Token } from "../types/data";
-import { MetricEvent, PageEvent, PingEvent, SummaryEvent, TagEvent, UploadEvent } from "../types/decode/data";
+import { MetricEvent, PageEvent, PingEvent, SummaryEvent, TagEvent, TargetEvent, UploadEvent } from "../types/decode/data";
 import { DecodedEvent, DecodedPayload } from "../types/decode/decode";
 import { ImageErrorEvent, ScriptErrorEvent } from "../types/decode/diagnostic";
 import { InputChangeEvent, PointerEvent, ResizeEvent, ScrollEvent } from "../types/decode/interaction";
 import { SelectionEvent, UnloadEvent, VisibilityEvent } from "../types/decode/interaction";
-import { BoxModelEvent, DocumentEvent, DomEvent, HashEvent, ResourceEvent, TargetEvent } from "../types/decode/layout";
+import { BoxModelEvent, DocumentEvent, DomEvent, HashEvent, ResourceEvent } from "../types/decode/layout";
 
 import * as data from "./data";
 import * as diagnostic from "./diagnostic";
@@ -44,6 +44,10 @@ export function decode(input: string): DecodedPayload {
             case Event.Tag:
                 if (payload.tag === undefined) { payload.tag = []; }
                 payload.tag.push(data.decode(entry) as TagEvent);
+                break;
+            case Event.Target:
+                if (payload.target === undefined) { payload.target = []; }
+                payload.target.push(data.decode(entry) as TargetEvent);
                 break;
             case Event.Metric:
                 if (payload.metric === undefined) { payload.metric = []; }
@@ -90,10 +94,6 @@ export function decode(input: string): DecodedPayload {
             case Event.Visibility:
                 if (payload.visibility === undefined) { payload.visibility = []; }
                 payload.visibility.push(interaction.decode(entry) as VisibilityEvent);
-                break;
-            case Event.Target:
-                if (payload.target === undefined) { payload.target = []; }
-                payload.target.push(layout.decode(entry) as TargetEvent);
                 break;
             case Event.BoxModel:
                 if (payload.boxmodel === undefined) { payload.boxmodel = []; }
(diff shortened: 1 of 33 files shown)


decode/clarity.ts
```diff
@@ -45,6 +45,10 @@ export function decode(input: string): DecodedPayload {
                 if (payload.tag === undefined) { payload.tag = []; }
                 payload.tag.push(data.decode(entry) as TagEvent);
                 break;
+            case Event.Target:
+                if (payload.target === undefined) { payload.target = []; }
+                payload.target.push(data.decode(entry) as TargetEvent);
```
toby-walker: >data [](start = 36, length = 4)

Why remove this case?

toby-walker:
>                 box: node && node.nodeType !== Node.TEXT_NODE ? layout(node, x, y) : []

Could you use null here instead of empty array.

---
Refers to: src/data/target.ts:34 in 875f412. [](commit_id = 875f41220e45c513307b4fb120749555309ace4a, deletion_comment = False)

src/interaction/change.ts
```diff
@@ -16,7 +15,6 @@ function recompute(evt: UIEvent): void {
     let input = evt.target as HTMLInputElement;
     let value = get(input);
     if (input && value) {
-        target.observe(value.id);
```
toby-walker: Why delete the target.observe? Wouldn't this miss sending in the target event?

src/interaction/encode.ts
```diff
@@ -73,7 +74,7 @@ export default function(type: Event): void {
             for (let i = 0; i < scroll.data.length; i++) {
                 let entry = scroll.data[i];
                 tokens = [entry.time, type];
-                tokens.push(entry.target);
+                tokens.push(observe(entry.target));
```
toby-walker: Is there any chance the new strategy of looking up targets at encoding time, rather than even processing time, will cause issue because of the latency and state not being the same?

src/interaction/scroll.ts
```diff
@@ -20,9 +19,11 @@ function recompute(event: UIEvent = null): void {
     let x = Math.round((eventTarget as HTMLElement).scrollLeft);
     let y = Math.round((eventTarget as HTMLElement).scrollTop);
     let id = getId(eventTarget as Node);
-    target.observe(id);
     let current: ScrollData = {target: id, x, y, time: time()};
 
+    // Bail out if we are attempting to retrieve initial scroll position and it turns out to be (0,0)
```
toby-walker: Comment isn't clear. Maybe something like 

We don't send any scroll events if the current position is at the top (0, 0). 

Also, what happen if the user scroll down and then back to 0,0. Wouldn't this change cause us to miss that case? I think this might need to be stateful -- on first event don't send 0,0 but send after that.

decode/clarity.ts
sarveshnagpal: Not removing it. Moving it up and handling it differently now that it's part of data.

src/interaction/encode.ts
sarveshnagpal: The logic doesn't change. We were still batching at the end. Earlier we had race conditions where targets could be part of a payload before or after. And with this change it ensures target and corresponding events are tightly coupled in same payload. There's always a risk that the event changes the dimensions of the box, however, that risk still exists if we attempt to get box immediately after. What would be ideal is if we got the dimensions of box *before* the event, which is unfortunately not possible / efficient. The risk of latency also hasn't changed from the previous implementation. There's a risk because here we are forced to process this synchronously, but we may revisit this as we gather more data. I can introduce complexity by adding a mix of async and sync fallback but I chose to keep it simple for now.

src/interaction/scroll.ts
sarveshnagpal: I'll rewrite the comment. But the state is ensured by checking for null event. That will only be possible when we call it from the start function. In all other subsequent calls, event will not be null and therefore won't hit this clause.

decode/clarity.ts
```diff
@@ -20,7 +20,7 @@ export function decode(input: string): DecodedPayload {
     let envelope = data.envelope(json.e);
     let timestamp = Date.now();
     let payload: DecodedPayload = { timestamp, envelope };
-    let encoded: Token[][] = json.d;
+    let encoded: Token[][] = json.d.sort((a: Token[], b: Token[]) => (a[0] as number) - (b[0] as number));
```
toby-walker: >Token [](start = 45, length = 5)

Add comment why you do this.

src/interaction/scroll.ts
```diff
@@ -20,9 +19,12 @@ function recompute(event: UIEvent = null): void {
     let x = Math.round((eventTarget as HTMLElement).scrollLeft);
     let y = Math.round((eventTarget as HTMLElement).scrollTop);
     let id = getId(eventTarget as Node);
-    target.observe(id);
     let current: ScrollData = {target: id, x, y, time: time()};
 
+    // We don't send any scroll events if the current position is at the top (0, 0)
```
toby-walker: >current [](start = 46, length = 7)

Reword to "We don't send any scroll events if this is the first event and the current position is top (0,0).

decode/clarity.ts
sarveshnagpal: Sure, done.

src/interaction/scroll.ts
sarveshnagpal: Done

PR #158, opened 2019-11-08
Improvements and bug fixes in v1.0.0-beta.12

sarveshnagpal:
* Changing the logic to compute selectors
* Updating metrics (coming soon)
* Any data quality issues from v11


decode/clarity.ts
```diff
@@ -52,7 +52,11 @@ export function decode(input: string): DecodedPayload {
                 break;
             case Event.Metric:
                 if (payload.metric === undefined) { payload.metric = []; }
-                payload.metric.push(data.decode(entry) as MetricEvent);
+                let metric = data.decode(entry) as MetricEvent;
```
toby-walker: Could you change the type of the key from a string to an int. Minor but it adds one more unnecessary change to the back end (the metric id is a integer).

decode/render.ts
```diff
@@ -8,27 +8,18 @@ let boxmodels = {};
...
 let lean = false;
 const METRIC_MAP = {};
-METRIC_MAP[Metric.Nodes] = { name: "Node Count", unit: ""};
+METRIC_MAP[Metric.Bytes] = { name: "Bytes", unit: "KB"};
 METRIC_MAP[Metric.LayoutBytes] = { name: "Layout Bytes", unit: "KB"};
 METRIC_MAP[Metric.InteractionBytes] = { name: "Interaction Bytes", unit: "KB"};
 METRIC_MAP[Metric.NetworkBytes] = { name: "Network Bytes", unit: "KB"};
-METRIC_MAP[Metric.DiagnosticBytes] = { name: "Diagnostic Bytes", unit: "KB"};
-METRIC_MAP[Metric.Mutations] = { name: "Mutation Count", unit: ""};
-METRIC_MAP[Metric.Interactions] = { name: "Interaction Count", unit: ""};
```
toby-walker: >Interactions [](start = 18, length = 12)

Worth keeping? Your call.

src/core/config.ts
```diff
@@ -2,7 +2,7 @@ import { Config } from "@clarity-types/core";
 
 let config: Config = {
     projectId: null,
-    longtask: 30, // 30 milliseconds
+    yield: 30, // 30 milliseconds
```
toby-walker: >yield [](start = 4, length = 5)

Might want to avoid keyboards. yieldTime?

src/core/task.ts
```diff
@@ -20,7 +19,7 @@ export async function schedule(task: TaskFunction): Promise<void> {
         queue.push({task, resolve});
     });
 
-    if (active === null) { run(); }
+    if (active === null) { requestAnimationFrame(run); }
```
toby-walker: >requestAnimationFrame [](start = 27, length = 21)

requestIdleCallback would be better? If not add a comment explaining why the requestAnimationFrame, which could block the UI thread with higher priority.

src/core/timeout.ts
```diff
@@ -0,0 +1,10 @@
+import { Event } from "@clarity-types/data";
+import wrap from "./wrap";
+
+export function setTimeout(handler: (event?: Event | boolean) => void, timeout: number, event?: Event): number {
+    return window.setTimeout(wrap(handler), timeout, event);
```
toby-walker: >setTimeout [](start = 18, length = 10)

Use a requestIdleCallback if available?

src/core/wrap.ts
```diff
@@ -0,0 +1,18 @@
+import { Metric } from "@clarity-types/data";
+import config from "@src/core/config";
+import * as metric from "@src/data/metric";
+
+// tslint:disable-next-line: ban-types
+export default function(method: Function): Function {
+    return function(): void {
+        let start = Date.now();
+        method.apply(this, arguments);
+        let duration = Date.now() - start;
```
toby-walker: Why not use perforamnce.now() for timing? More precise.

src/core/wrap.ts
```diff
@@ -0,0 +1,18 @@
...
+        metric.counter(Metric.Cost, duration);
```
toby-walker: >counter [](start = 15, length = 7)

Doesn't seem like a counter to me. :)

src/core/wrap.ts
```diff
@@ -0,0 +1,18 @@
...
+        metric.counter(Metric.Calls);
+        if (duration > config.yield) {
+            metric.counter(Metric.LongTasks);
+            metric.measure(Metric.MaxBlockTime, duration);
```
toby-walker: >MaxBlockTime [](start = 34, length = 12)

Should this not be a max(duration and previous value)?

src/data/index.ts
```diff
@@ -7,19 +8,19 @@ import * as upload from "@src/data/upload";
...
-    tag.reset();
-    ping.end();
-    upload.end();
-    target.reset();
-    metadata.end();
+    wrap(tag.reset)();
+    wrap(ping.end)();
+    wrap(upload.end)();
+    wrap(target.reset)();
+    wrap(metadata.end)();
```
toby-walker: >wrap [](start = 4, length = 4)

Minor; wrap is very general and doesn't tell you what it is doing. Why is it wrapped and what does the wrapper do? . Maybe time or measure or something a bit more specific.

src/layout/boxmodel.ts
```diff
@@ -11,23 +12,23 @@ let timeout: number = null;
...
+    timeout = setTimeout(schedule, config.lookahead);
 }
 
 function schedule(): void {
     task.schedule(boxmodel);
 }
 
 async function boxmodel(): Promise<void> {
-    let timer = Metric.BoxModelTime;
+    let timer = Metric.BoxModelCost;
```
toby-walker: >BoxModelCost [](start = 23, length = 12)

Should come up with a naming convention for metric ids including: 

* xxxCount coutns something
* xxxDuration. Counts the time of something in known units (e.g., milliseconds)
* xxxTime. The time of something measured on the performance now timeline in ms.
* xxxTimestamp. Full epoch timestamp value (not used)?
* xxxByteCount of xxxBytes or xxxSize. Number of bytes in something.

decode/clarity.ts
sarveshnagpal: It is an integer. I think you may receive it as a string because by default JSON.stringify (native function) treats keys of an object as string. There's not much I can do without overloading JSON.stringify, which I would want to avoid to keep things as native as possible. Nothing should change for you though from previous version - it's still the same schema.

decode/render.ts
sarveshnagpal: Interactions was also a count of events like: Scroll, Pointer, Selection, etc. I think based on our offline thread, you could easily compute it with the events we have. It essentially is a count of events.

src/core/config.ts
sarveshnagpal: Good point. Let me go back to longtask as the setting name.

src/core/task.ts
sarveshnagpal: There are similar concepts and agreed requestIdleCallback would be ideal but Edge doesn't support requestIdleCallback yet: https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback. Instead of getting into pollyfills - I stuck with a stable parallel API. As far as I understand, requestAnimationFrame is essentially gives us an opportunity to execute at the next time UI thread is free. It wouldn't block the thread. It is also consistent with what we were using in wait function in the same file.

src/core/timeout.ts
sarveshnagpal: Not adding extra complexity since not all modern browsers support requestIdleCallback. It's still an experimental technology and pollyfills rely on setTimeout / requestAnimationFrame which we are using. We can revisit using requestIdleCallback when it's more mainstream. Also, I am not sure if that's what we really want based on the descript. We can chat more offline and talk about it future versions.

src/core/wrap.ts
sarveshnagpal: Please suggest a better name and I am happy to change the name :)

src/core/wrap.ts
sarveshnagpal: metric.measure takes care of max(duration, previous). we don't have to do that at call time.

src/data/index.ts
sarveshnagpal: I kept it generic on purpose because eventually we may even want to introduce try / catch to eat exceptions if they happen. I looked at few other APIs and they go with generic wrap - so they can do few things at once. If you feel strongly about it, I will change the name. Please let me know.

src/layout/boxmodel.ts
sarveshnagpal: Good point. I will follow these conventions. Thanks.

src/core/wrap.ts
sarveshnagpal: Refactored per your suggestion in the last comment. Called it duration now.

src/data/index.ts
sarveshnagpal: Renamed it to measure for now. When we need to do more, we will come back and call it wrap or something else later.

src/layout/boxmodel.ts
sarveshnagpal: Followed this convention. Except for cost, I used Latency as the suffix: xxxLatency: (ms)

src/core/wrap.ts
sarveshnagpal: Good point. Using performance.now() now and rounding off at the time of encoding so we don't have to deal with floating points while uploading.

src/core/task.ts
toby-walker: I don't think you need polyfills. Just encapsulate it like you did setTimer and then you can change it later.

---
In reply to: [344512522](https://github.com/microsoft/clarity-js/pull/158#discussion_r344512522) [](ancestors = 344512522)

toby-walker: Right, but requestIdelCallback gets a calback when the browser is "idle", which is not the same as for requestAnimationCallback (which is supposed to be tied to the frame rate of the rendering so that you get one call per frame. 

---
In reply to: [344513493](https://github.com/microsoft/clarity-js/pull/158#discussion_r344513493) [](ancestors = 344513493)

decode/render.ts
```diff
@@ -8,27 +8,18 @@ let boxmodels = {};
...
-METRIC_MAP[Metric.ActiveTime] = { name: "Active Time", unit: "ms"};
-METRIC_MAP[Metric.EndTime] = { name: "End Time", unit: "s"};
-METRIC_MAP[Metric.ViewportWidth] = { name: "Viewport Width", unit: "px"};
-METRIC_MAP[Metric.ViewportHeight] = { name: "Viewport Height", unit: "px"};
-METRIC_MAP[Metric.DocumentWidth] = { name: "Document Width", unit: "px"};
-METRIC_MAP[Metric.DocumentHeight] = { name: "Document Height", unit: "px"};
+METRIC_MAP[Metric.TargetBytes] = { name: "Target Bytes", unit: "KB"};
+METRIC_MAP[Metric.InvokeCount] = { name: "Invoke Count" };
+METRIC_MAP[Metric.LongTaskCount] = { name: "Long Tasks" };
+METRIC_MAP[Metric.Latency] = { name: "Latency", unit: "ms"};
```
toby-walker: >Latency [](start = 18, length = 7)

This is very broad. The latency of what? Can you add a prefix?

decode/render.ts
```diff
@@ -8,27 +8,18 @@ let boxmodels = {};
...
+METRIC_MAP[Metric.DiscoverLatency] = { name: "Discover Latency", unit: "ms"};
+METRIC_MAP[Metric.MutationLatency] = { name: "Mutation Latency", unit: "ms"};
+METRIC_MAP[Metric.BoxModelLatency] = { name: "BoxModel Latency", unit: "ms"};
+METRIC_MAP[Metric.ThreadBlockDuration] = { name: "Thread Block", unit: "ms"};
```
toby-walker: >ThreadBlockDuration [](start = 18, length = 19)

What's the diff between duration and latency above?

src/core/measure.ts
```diff
@@ -0,0 +1,18 @@
...
+// tslint:disable-next-line: ban-types
+export default function(method: Function): Function {
+    return function(): void {
+        let start = performance.now();
+        method.apply(this, arguments);
+        let duration = performance.now() - start;
+        metric.duration(Metric.Latency, duration);
+        metric.counter(Metric.InvokeCount);
+        if (duration > config.longtask) {
+            metric.counter(Metric.LongTaskCount);
```
toby-walker: >counter [](start = 19, length = 7)

Minor but I prefer verb to noun names so would change these to metric.count(LongTaskCount).

src/data/encode.ts
```diff
@@ -53,7 +52,7 @@ export default function(event: Event): void {
                         let m = parseInt(d, 10);
                         if (metric.updates.indexOf(m) >= 0) {
                             tokens.push(m);
-                            tokens.push(metric.data[d]);
+                            tokens.push(Math.round(metric.data[d]));
```
toby-walker: What if value can be floating point? Add comment as this is cryptic here.

src/data/metric.ts
```diff
@@ -18,7 +18,13 @@ export function counter(metric: Metric, increment: number = 1): void {
     track(metric);
 }
 
-export function measure(metric: Metric, value: number): void {
+export function duration(metric: Metric, time: number): void {
```
toby-walker: >duration [](start = 16, length = 8)

Suggest accumulate as better name for this.

decode/render.ts
sarveshnagpal: Need a way to differentiate cumulative sum (latency) and something that just captures the max (not cumulative). Any suggestions on the naming?

src/core/measure.ts
sarveshnagpal: Sure, will make the change.

decode/render.ts
sarveshnagpal: Called it TotalLatency and TotalBytes.

src/data/encode.ts
sarveshnagpal: Done

src/data/metric.ts
sarveshnagpal: Done

decode/render.ts
sarveshnagpal: Resolved this offline. Changes reflected in updated PR.

PR #159, opened 2019-11-15
Data quality fixes in v1.0.0-beta.13

sarveshnagpal:
Synchronizing tasks in the right order to improve data quality. This change updates interactions with dependency on layout (e.g. target element) to be processed in the right order so we do not hit target is null data quality issue. This change also exposes reset from decode module for better control over visualization.

### DIFF
--- a/decode/render.ts
+++ b/decode/render.ts
@@ -20,9 +20,9 @@ METRIC_MAP[Metric.DiscoverDuration] = { name: "Discover", unit: "ms"};
 METRIC_MAP[Metric.MutationDuration] = { name: "Mutation", unit: "ms"};
 METRIC_MAP[Metric.BoxModelDuration] = { name: "BoxModel", unit: "ms"};
 METRIC_MAP[Metric.MaxThreadBlockedDuration] = { name: "Thread Blocked", unit: "ms"};
-METRIC_MAP[Metric.DataDuration] = { name: "Data Duration", unit: "ms"};
-METRIC_MAP[Metric.DiagnosticDuration] = { name: "Diagnostic Duration", unit: "ms"};
-METRIC_MAP[Metric.InteractionDuration] = { name: "Interaction Duration", unit: "ms"};
+METRIC_MAP[Metric.DataDuration] = { name: "Data", unit: "ms"};
+METRIC_MAP[Metric.DiagnosticDuration] = { name: "Diagnostic", unit: "ms"};
+METRIC_MAP[Metric.InteractionDuration] = { name: "Interaction", unit: "ms"};
 
 // tslint:disable-next-line: max-line-length
 let pointerIcon = "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABQAAAAYCAYAAAD6S912AAAABGdBTUEAALGOfPtRkwAAACBjSFJNAAB6JQAAgIMAAPn/AACA6AAAdTAAAOpgAAA6lwAAF2+XqZnUAAACaUlEQVR4nGL8f58BHYgAsT8Q2wGxBBAzQcX/AfFrID4CxOuA+BWKLoX/YAoggBjRDHQD4ngglgRiPgyrIOAzEL8E4lVQg1EMBAggFiSFYUAcA8RSOAyCAV4oTgViTiBeiiwJEEAw71gRaRgyEAXiKCB2RBYECCCQgcIMEG+SYhgMiANxEhDzwwQAAghkoAMQK5NhGAwoALE1jAMQQCADQU7mpMBAZqijwAAggEAGqgAxOwUGskHNAAOAAAIZyEtIh4INg3bfHHD6xAUEYAyAAAIZ+IuQgU9fMLCXdzDIzV3JIIhDyQ8YAyCAQAaCUv8/fAZysDP8+/OXgTG7jkFhwRoMQ0F6n8M4AAEEMvAKsg34wM9fDEwgQ1dtRSQTIPgNxFdhHIAAAhm4AYg/EmMgCHz7zsCUVMaguHob3FCQYzbD5AECCGTgJSDeCbWJKPD1GwNzSjmD4tZ9DFxgvQr/b8PkAAIIlvVWA/FuUgz99IWBOTyXQcE+nOEOsjhAACGXNnJAHAnE9kAshqyIV5vB4Ms3cALGBkAlj9////9PgTgAAcSEJPEIiDuBeBYQP2CAhOt3BsLJCpSfNzAyMpqDOAABhF4ewh3FAMmf2kAsyqnBUPDjJ8HcdBvoSjWAAGIEEgTUMTAAbf/AwICSVGCgD4hPgJQA8WegWdsBAogFiyJC4C0QgxI3KLj4gIasRpYECCAGkAsJYSAAuRDEAKUEQwZIzgDxvwCxCrJagAAi1kAQAYpFESh/BlQMhJuR1QIEELEGlgOxHBLflAGSh0Gc60DMBpMDCCCiDMRhyXoGSJUaDgpPmDhAgAEAN5Ugk0bMYNIAAAAASUVORK5CYII=";
@@ -228,7 +228,7 @@ function setAttributes(node: HTMLElement, attributes: object): void {
 }
 
 export function scroll(data: ScrollData, iframe: HTMLIFrameElement): void {
-    let target = getNode(data.target);
+    let target = getNode(data.target as number);
     if (target) { target.scrollTo(data.x, data.y); }
 }
 
@@ -254,14 +254,14 @@ export function resize(data: ResizeData, iframe: HTMLIFrameElement): void {
 }
 
 export function change(data: InputChangeData, iframe: HTMLIFrameElement): void {
-    let el = element(data.target) as HTMLInputElement;
+    let el = element(data.target as number) as HTMLInputElement;
     if (el) { el.value = data.value; }
 }
 
 export function selection(data: SelectionData, iframe: HTMLIFrameElement): void {
     let doc = iframe.contentDocument;
     let s = doc.getSelection();
-    s.setBaseAndExtent(element(data.start), data.startOffset, element(data.end), data.endOffset);
+    s.setBaseAndExtent(element(data.start as number), data.startOffset, element(data.end as number), data.endOffset);
 }
 
 export function pointer(event: Event, data: PointerData, iframe: HTMLIFrameElement): void {
--- a/decode/render.ts
+++ b/decode/render.ts
@@ -20,6 +20,9 @@ METRIC_MAP[Metric.DiscoverDuration] = { name: "Discover", unit: "ms"};
 METRIC_MAP[Metric.MutationDuration] = { name: "Mutation", unit: "ms"};
 METRIC_MAP[Metric.BoxModelDuration] = { name: "BoxModel", unit: "ms"};
 METRIC_MAP[Metric.MaxThreadBlockedDuration] = { name: "Thread Blocked", unit: "ms"};
+METRIC_MAP[Metric.DataDuration] = { name: "Data Duration", unit: "ms"};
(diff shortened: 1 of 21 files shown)


toby-walker:
> import * as r from "./render";

Nit: r is very cryptic. Can you just use "render"?

---
Refers to: decode/clarity.ts:14 in 1d719ff. [](commit_id = 1d719ffd1f5b81207fa09c9d11d49a00eee8f3d3, deletion_comment = False)

decode/render.ts
```diff
@@ -251,14 +254,14 @@ export function resize(data: ResizeData, iframe: HTMLIFrameElement): void {
 }
 
 export function change(data: InputChangeData, iframe: HTMLIFrameElement): void {
-    let el = element(data.target) as HTMLInputElement;
+    let el = element(data.target as number) as HTMLInputElement;
```
toby-walker: >target [](start = 26, length = 6)

Nit: Why do you need this override? Shouldn't the be in the Typescript?

decode/render.ts
```diff
@@ -251,14 +254,14 @@ export function resize(data: ResizeData, iframe: HTMLIFrameElement): void {
...
     if (el) { el.value = data.value; }
 }
 
 export function selection(data: SelectionData, iframe: HTMLIFrameElement): void {
```
toby-walker: There's a design issue here. We really need not the start/end but the set of element in the selection. For a future PR.

src/core/task.ts
```diff
@@ -1,9 +1,10 @@
-import { AsyncTask, TaskFunction, TaskResolve, TaskTiming } from "@clarity-types/core";
+import { AsyncTask, TaskFunction, TaskResolve } from "@clarity-types/core";
 import { Metric } from "@clarity-types/data";
 import config from "@src/core/config";
 import * as metric from "@src/data/metric";
 
-let tracker: TaskTiming = {};
+let tracker: { [key: number]: number } = {};
```
toby-walker: Please add a comment about wat these represent. From names and data type very hard to glean much.

src/core/task.ts
```diff
@@ -19,7 +20,9 @@ export async function schedule(task: TaskFunction): Promise<void> {
         queue.push({task, resolve});
     });
 
-    if (active === null) { requestAnimationFrame(run); }
+    // If the task queue is empty, invoke the first task in the queue synchronously
+    // This also ensures we don't yield the thread during unload event
+    if (active === null) { run(); }
```
toby-walker: >null [](start = 19, length = 4)

Why does active===null mean the same things as queue.lenght == 0? From the comment the latter condition is what I would expect you check.

src/core/task.ts
```diff
@@ -41,22 +44,26 @@ export function blocking(method: Metric): boolean {
     return (elapsed > config.longtask);
 }
 
-export function start(method: Metric): void {
+export function start(method: Metric, resume: boolean = false): void {
```
toby-walker: You don't use the value. Why is it needed?

src/core/task.ts
```diff
@@ -41,22 +44,26 @@ export function blocking(method: Metric): boolean {
...
     tracker[method] = performance.now();
+    counter[method] = resume ? counter[method] + 1 : 0;
 }
 
-export function stop(method: Metric): void {
+export function stop(method: Metric, pause: boolean = false): void {
```
toby-walker: Not used. Why?

toby-walker:
>             active = null;

This comment needs a *lot* more explanation.

---
Refers to: src/core/task.ts:36 in 1d719ff. [](commit_id = 1d719ffd1f5b81207fa09c9d11d49a00eee8f3d3, deletion_comment = False)

src/data/target.ts
```diff
@@ -10,15 +10,20 @@ export function reset(): void {
...
-            box: node && node.nodeType !== Node.TEXT_NODE ? layout(node) : null
-        };
+export function observe(node: Node): number {
+    let id = null;
+    if (node !== null) {
+        let value = dom.get(node);
+        if (value !== null) {
+            id = value.id;
+            if (id !== null && !(id in queue)) {
+                queue[id] = {
```
toby-walker: How does the value get "fixed up" in this case (i.e. when value.id === null? Or once it gets into the queue?

src/diagnostic/encode.ts
```diff
@@ -19,7 +19,7 @@ export default function(type: Event): Token[] {
             break;
         case Event.ImageError:
             tokens.push(image.data.source);
-            tokens.push(observe(image.data.target));
+            tokens.push(observe(image.data.target as Node));
```
toby-walker: >observe [](start = 24, length = 7)

How does this get fixed up if the target is not known yet and the observe function returns null? Put the question here but it is a generic one for the entire approach.

decode/render.ts
sarveshnagpal: This one's tricky and I've gone back and forth on this. The fact that we have to hold target element as a Node until very late when we are ready to turn it into an id to send over the wire makes it hard. The alternate option would be to fork all interfaces that have target and introduce two fields - one for target Node and another for target Id. I started going down that road and felt we ended up making our interfaces confusing. With this change target can have two possible types (which is defined in data.d.ts) - Node or number. So we cast it in the right format. My preference is what I've in the PR. Let me know if you feel strongly about it and we can chat offline about this. It's probably hard to cover all the issues I encountered over PR.

decode/render.ts
sarveshnagpal: Yep, I see that. I'll fix it in next version where I am going to introduce Performance events and will have some schema changes.

src/core/task.ts
sarveshnagpal: Done

src/diagnostic/image.ts
```diff
@@ -13,10 +13,7 @@ export function start(): void {
 function handler(error: ErrorEvent): void {
     let target = error.target as HTMLElement;
     if (target && target.tagName === "IMG") {
-        data = {
-            source: (target as HTMLImageElement).src,
-            target: getId(target)
-        };
-        encode(Event.ImageError);
+        data = { source: (target as HTMLImageElement).src, target };
+        schedule(encode.bind(this, Event.ImageError));
```
toby-walker: >schedule [](start = 8, length = 8)

Wouldn't it be clearer to actually track the promises for these by element id and then, when it resolves, invoke them?

src/interaction/encode.ts
```diff
@@ -25,12 +28,13 @@ export default function(type: Event): void {
...
         case Event.TouchMove:
         case Event.TouchCancel:
-            for (let i = 0; i < pointer.data[type].length; i++) {
-                let entry = pointer.data[type][i];
-                tokens = [entry.time, type];
-                tokens.push(observe(entry.target));
-                tokens.push(entry.x);
-                tokens.push(entry.y);
+            for (let i = 0; i < pointer.data.length; i++) {
+                if (task.blocking(timer)) { await task.idle(timer); }
```
toby-walker: >blocking [](start = 25, length = 8)

What are you doing in this code (the handler for touch cancel)? Why is this different from all other events?

toby-walker:
> export function blocking(method: Metric): boolean {

This function name is confusng to me. Wouldn't something like "shouldYield" be *much* clearer. 

---
Refers to: src/core/task.ts:42 in 1d719ff. [](commit_id = 1d719ffd1f5b81207fa09c9d11d49a00eee8f3d3, deletion_comment = False)

src/core/task.ts
sarveshnagpal: The comment was confusing. I've updated it to reflect that no active task is running. Queue won't be empty at this time because just inserted a task in the queue. Fixed the comment.

src/core/task.ts
sarveshnagpal: I left it for consistency to be used in future since I added "resume" to start. If you feel this is redundant, I can remove it for now and add it back when we need it for any reason.

src/interaction/encode.ts
```diff
@@ -50,27 +54,28 @@ export default function(type: Event): void {
...
             tokens.push(s.startOffset);
-            tokens.push(observe(s.end));
+            tokens.push(observe(s.end as Node));
             tokens.push(s.endOffset);
             queue(tokens);
             selection.reset();
             break;
         case Event.Scroll:
             for (let i = 0; i < scroll.data.length; i++) {
+                if (task.blocking(timer)) { await task.idle(timer); }
```
toby-walker: >idle [](start = 55, length = 4)

Woldn't this be better call "yield" or similar? I think that's what is doing. Idel is a state and this is an action. Notice I have a general preference for "verb-sh" function names. You seems to prefer noun-ish ones.

src/interaction/pointer.ts
```diff
@@ -28,64 +28,64 @@ function mouse(event: Event, evt: MouseEvent): void {
...
     event = event === Event.Click && (evt.buttons === 2 || evt.button === 2) ? Event.RightClick : event;
-    handler(event, {target: id, x, y, time: time()});
+    handler({ time: time(), event, data: { target, x, y } });
 }
 
 function touch(event: Event, evt: TouchEvent): void {
     let de = document.documentElement;
     let touches = evt.changedTouches;
-    let id = evt.target ? getId(evt.target as Node) : null;
+    let target = evt.target ? evt.target as Node : null;
```
toby-walker: >null [](start = 51, length = 4)

When can this case occur?

src/data/target.ts
sarveshnagpal: The assumption here is that we will get to "observe" only after all previous DOM operations have completed (discover / mutations). So when we come here, the value should be present - of course we will know more from the real data that comes back. We will receive null value only for unexpected cases at this time, e.g. elements that we are not tracking for some reason but still got in as target element.

src/interaction/pointer.ts
toby-walker: >null [](start = 51, length = 4)

Why can this case occur?

src/interaction/pointer.ts
```diff
@@ -28,64 +28,64 @@ function mouse(event: Event, evt: MouseEvent): void {
...
-            if (last && similar(last, current)) { data[event].pop(); }
-            data[event].push(current);
+            let length = data.length;
+            let last = length > 1 ? data[length - 2] : null;
+            if (last && similar(last, current)) { data.pop(); }
+            data.push(current);
 
             clearTimeout(timeout);
-            timeout = setTimeout(encode, config.lookahead, event);
+            timeout = setTimeout(process, config.lookahead, current.event);
```
toby-walker: Would be clearer to me if you just passed in a lambda to this.

toby-walker: Minor: This would be clearer if you passed in just a lambda function and didn't pass the fn and argument(s) separately.

src/interaction/pointer.ts
```diff
@@ -28,64 +28,64 @@ function mouse(event: Event, evt: MouseEvent): void {
...
-    return (distance < config.distance) && (current.time - last.time < config.interval) && current.target === last.target;
+    let gap = current.time - last.time;
+    return current.event === last.event && distance < config.distance && gap < config.interval && current.data.target === last.data.target;
 }
 
 export function end(): void {
     clearTimeout(timeout);
-    data = {};
+    // Send out any pending pointer events in the pipeline
+    if (data.length > 0) { process(data[data.length - 1].event); }
```
toby-walker: Isn't this needed for all other events that use the new task framework?

types/diagnostic.d.ts
```diff
@@ -10,5 +11,5 @@ export interface ScriptErrorData {
 
 export interface ImageErrorData {
     source: string;
-    target: number;
+    target: Target;
```
toby-walker: >Target [](start = 12, length = 6)

This is mixing the typing for the wire data and the in-memory representation. Ideally you'd separate them into two different types if they don't agree.

For now you could at least make this number | Target and add a comment why the union type is needed.

src/interaction/encode.ts
toby-walker: Same comment above as for touch cancel. Why do you have to do this yielding logic only for this event?

PR #160, opened 2019-11-18
Bug fixes and support for performance observer in v1.0.0-beta.14

sarveshnagpal:
* Fixing script error when getBoundingClientRect may not be available on a node
* Keeping *B around for visualization to be able to pickup base href
* Fixing scroll computations on Edge browser
* Clearing up pending tasks on teardown
* Adding support for PerformanceObserver, NetInfo and performance.memory API.


decode/clarity.ts
```diff
@@ -55,6 +58,8 @@ export function decode(input: string): DecodedPayload {
                 let metric = data.decode(entry) as MetricEvent;
                 // It's not possible to accurately include the byte count of the payload within the same payload
                 // So, we increment the bytes from the incoming payload at decode time.
+                // Alos, initialize TotalBytes if it doesn't exist. For the first payload, this value can be null.
```
ender336: >Alos [](start = 19, length = 4)

also

src/layout/boxmodel.ts
```diff
@@ -71,22 +71,24 @@ function update(id: number, box: number[]): void {
 
 export function layout(element: Element, x: number = null, y: number = null): number[] {
     let box: number[] = null;
-    let rect = element.getBoundingClientRect();
-    x = x !== null ? x : ("pageXOffset" in window ? window.pageXOffset : document.documentElement.scrollLeft);
-    y = y !== null ? y : ("pageYOffset" in window ? window.pageYOffset : document.documentElement.scrollTop);
+    if (typeof element.getBoundingClientRect === "function") {
```
ender336: >getBoundingClientRect [](start = 23, length = 21)

is any of the boxmodel code worth running without this? wondering if initial wireup should check for getBoundingClientRect and not enable boxmodel if it doesnt exist

decode/performance.ts
```diff
@@ -0,0 +1,105 @@
...
+    switch (event) {
+        case Event.Connection:
+            let connectionData: ConnectionData = {
+                downlink: tokens[2] as number,
+                rtt: tokens[3] as number,
+                saveData: tokens[4] as number,
+                type: tokens[5] as string
+            };
+            return { time, event, data: connectionData };
+        case Event.Contentful:
```
toby-walker: >Contentful [](start = 19, length = 10)

Is this for a ContentfulPaint? I'd keep the full name as the shortened form doesn't make sense.

decode/performance.ts
```diff
@@ -0,0 +1,105 @@
...
+            let network = [];
+            let stringIndex = 0;
+            let networkData: NetworkData[] = [];
+            for (let i = 2; i < tokens.length; i++) {
+                let token = tokens[i];
+                let type = typeof(token);
+                switch (type) {
+                    case "number":
+                        if (type !== lastType && lastType !== null) {
+                            networkData.push(process(network, stringIndex));
```
toby-walker: >process [](start = 45, length = 7)

This is pretty messy. Can you explain what it's doing and put a few comment as signposts?

src/data/token.ts
```diff
@@ -1,10 +1,18 @@
...
+import {Token} from "@clarity-types/data";
 
-export function check(hash: string): boolean {
-    let output = tokens.indexOf(hash) >= 0;
-    return output;
-}
-
-export function resolve(hash: string): string[] {
-    return check(hash) ? tokens[hash] : [];
+export default function(tokens: Token[], metadata: Token[]): Token[] {
```
toby-walker: What is this doing? Add some comments. Code is cryptic.

src/layout/boxmodel.ts
toby-walker: +1. Also can we send a signal that this situation happened so we can tell that the data is missing "by design" because the browser doesn't support it?

---
In reply to: [347530487](https://github.com/microsoft/clarity-js/pull/160#discussion_r347530487) [](ancestors = 347530487)

src/layout/dom.ts
```diff
@@ -217,16 +225,18 @@ function remove(id: number, source: Source): void {
     value.children = [];
 }
 
-function layout(tag: string, id: number, parentId: number): void {
+function meta(tag: string, id: number, parentId: number): void {
```
toby-walker: What does "meta" mean? Earlier name seemed clearer to me.

src/layout/dom.ts
```diff
@@ -235,17 +245,47 @@ function layout(tag: string, id: number, parentId: number): void {
                 }
                 break;
             case "IMG":
+                value.metadata.boxmodel = true;
+            case "VIDEO":
+            case "AUDIO":
+            case "LINK":
+                // Track mapping between URL and corresponding nodes
+                if (Constant.HREF_ATTRIBUTE in attributes && attributes[Constant.HREF_ATTRIBUTE].length > 0) {
+                    urlMap[getFullUrl(attributes[Constant.HREF_ATTRIBUTE])] = id;
```
toby-walker: What if same url maps to multiple ids?

src/performance/network.ts
```diff
@@ -0,0 +1,46 @@
...
+import { NetworkState } from "@clarity-types/performance";
+import config from "@src/core/config";
+import { schedule } from "@src/core/task";
+import { clearTimeout, setTimeout } from "@src/core/timeout";
+import encode from "./encode";
+
+export let state: NetworkState[] = [];
+let timeout: number = null;
+
+export function compute(entry: PerformanceResourceTiming): void {
```
toby-walker: >PerformanceResourceTiming [](start = 31, length = 25)

Small preference: Keep the resource name of the fail and event type from before. That keeps the naming alignment between "xxx Timing" and xxx consistent.

src/performance/network.ts
```diff
@@ -0,0 +1,46 @@
...
+    if (config.url.length > 0 && entry.name.indexOf(config.url) >= 0) { return; }
+
+    state.push({
+        url: entry.name,
+        data: {
+            start: Math.round(entry.startTime),
+            duration: Math.round(entry.duration),
+            size: entry.transferSize,
+            target: null,
+            initiator: entry.initiatorType,
```
toby-walker: Might want encoded/decoded size?

src/performance/navigation.ts
```diff
@@ -0,0 +1,27 @@
...
+        fetchStart: Math.round(entry.fetchStart),
+        connectStart: Math.round(entry.connectStart),
+        connectEnd: Math.round(entry.connectEnd),
+        requestStart: Math.round(entry.requestStart),
+        responseStart: Math.round(entry.responseStart),
+        responseEnd: Math.round(entry.responseEnd),
+        domInteractive: Math.round(entry.domInteractive),
+        domComplete: Math.round(entry.domComplete),
+        loadEventEnd: Math.round(entry.loadEventEnd),
+        size: entry.transferSize,
```
toby-walker: Add redirectCount

src/performance/navigation.ts
```diff
@@ -0,0 +1,27 @@
...
+    data = {
+        fetchStart: Math.round(entry.fetchStart),
+        connectStart: Math.round(entry.connectStart),
+        connectEnd: Math.round(entry.connectEnd),
+        requestStart: Math.round(entry.requestStart),
+        responseStart: Math.round(entry.responseStart),
+        responseEnd: Math.round(entry.responseEnd),
+        domInteractive: Math.round(entry.domInteractive),
+        domComplete: Math.round(entry.domComplete),
+        loadEventEnd: Math.round(entry.loadEventEnd),
```
toby-walker: >loadEventEnd [](start = 8, length = 12)

loadEndStart

src/performance/navigation.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export function compute(entry: PerformanceNavigationTiming): void {
+    data = {
+        fetchStart: Math.round(entry.fetchStart),
+        connectStart: Math.round(entry.connectStart),
+        connectEnd: Math.round(entry.connectEnd),
+        requestStart: Math.round(entry.requestStart),
+        responseStart: Math.round(entry.responseStart),
+        responseEnd: Math.round(entry.responseEnd),
+        domInteractive: Math.round(entry.domInteractive),
```
toby-walker: Add domContentLoadedEventStart;, domContentLoadedEventEnd;

src/performance/paint.ts
```diff
@@ -0,0 +1,14 @@
+import { Event } from "@clarity-types/data";
+import { PaintState } from "@clarity-types/performance";
+import encode from "./encode";
+
+export let state: PaintState = null;
```
toby-walker: For each of these performance entry events could you add a link to the w3c spec you based it on? Fore xampel, this one is:
[https://www.w3.org/TR/paint-timing/](https://www.w3.org/TR/paint-timing/)

src/performance/memory.ts
```diff
@@ -0,0 +1,21 @@
...
+import encode from "./encode";
+
+export let data: MemoryData = null;
+
+export function reset(): void {
+    data = null;
+}
+
+export function compute(): void {
+    if (performance && "memory" in performance) {
```
toby-walker: >memory [](start = 24, length = 6)

Add link to spec this is based on. Never heard of it. :-)

src/performance/longtask.ts
```diff
@@ -0,0 +1,14 @@
...
+import encode from "./encode";
+
+export let state: LongTaskState = null;
+
+export function reset(): void {
+    state = null;
+}
+
+export function compute(entry: PerformanceEntry): void {
+    state = { time: Math.round(entry.startTime), data: { duration: Math.round(entry.duration) } };
```
toby-walker: Add attribution property. See [https://w3c.github.io/longtasks/#sec-PerformanceLongTaskTiming](https://w3c.github.io/longtasks/#sec-PerformanceLongTaskTiming)

src/performance/connection.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export function start(): void {
+    if (navigator && "connection" in navigator) {
+        (navigator["connection"] as NavigatorConnection).addEventListener("change", recompute);
+        recompute();
+    }
+}
+
+function recompute(): void {
+    let connection = navigator["connection"] as NavigatorConnection;
```
toby-walker: >NavigatorConnection [](start = 48, length = 19)

Add comment with link to [http://wicg.github.io/netinfo/#dom-networkinformation](http://wicg.github.io/netinfo/#dom-networkinformation)

src/performance/connection.ts
```diff
@@ -0,0 +1,31 @@
...
+    data = {
+        downlink: connection.downlink,
```
toby-walker: Add downlinkMax,, type

src/performance/contentful.ts
```diff
@@ -0,0 +1,20 @@
+import { Event } from "@clarity-types/data";
+import { LargestContentfulData, LargestContentfulPaint } from "@clarity-types/performance";
+import { schedule } from "@src/core/task";
+import encode from "./encode";
+
+export let data: LargestContentfulData = null;
```
toby-walker: Add link to [https://wicg.github.io/largest-contentful-paint/](https://wicg.github.io/largest-contentful-paint/)
interface LargestContentfulPaint : PerformanceEntry {
    readonly attribute DOMHighResTimeStamp renderTime;
    readonly attribute DOMHighResTimeStamp loadTime;
    readonly attribute unsigned long size;
    readonly attribute DOMString id;
    readonly attribute DOMString url;
    readonly attribute Element? element;
    [Default] object toJSON();
};

decode/clarity.ts
sarveshnagpal: Thanks

---
In reply to: [347527567](https://github.com/microsoft/clarity-js/pull/160#discussion_r347527567) [](ancestors = 347527567)

src/layout/boxmodel.ts
sarveshnagpal: This is a fair point. What I have right now is more of a mitigation. One known case is when we have a text node instead of an element, and not so much to do with browser. I will tackle this in subsequent ones, together with selection event. We probably will have to walk up the DOM chain to get an element as target and not text nodes - which are hard to track in our current model (e.g. there's no selector).

decode/performance.ts
sarveshnagpal: Yes, will change. Thanks

src/layout/dom.ts
sarveshnagpal: The last one based on mutation will win. We have 1:1 right now, no support for 1 to many. Let me know if you have thoughts on how to change it for future version?

src/performance/network.ts
sarveshnagpal: In this case PerformanceResourceTiming is a native Typescript type from W3 specs. We will have to rename the type and redefine custom type. I will avoid adding custom types where possible. Please let me know if you think I missed something here.

---
In reply to: [347585636](https://github.com/microsoft/clarity-js/pull/160#discussion_r347585636) [](ancestors = 347585636)

src/performance/network.ts
sarveshnagpal: I kept it out intentionally but we can add more fields as the need arises. Even for full URL, I am only storing the host to not go overboard. Basically transferSize is super set of encoded size. Decode is more of after compression and wasn;t sure if it was immediately helpful. 

---
In reply to: [347586302](https://github.com/microsoft/clarity-js/pull/160#discussion_r347586302) [](ancestors = 347586302)

src/performance/navigation.ts
sarveshnagpal: I will add it, thanks.

---
In reply to: [347587052](https://github.com/microsoft/clarity-js/pull/160#discussion_r347587052) [](ancestors = 347587052)

src/performance/navigation.ts
sarveshnagpal: Done, added. Thanks.

---
In reply to: [347587271](https://github.com/microsoft/clarity-js/pull/160#discussion_r347587271) [](ancestors = 347587271)

src/performance/paint.ts
sarveshnagpal: Good point, adding.

---
In reply to: [347587865](https://github.com/microsoft/clarity-js/pull/160#discussion_r347587865) [](ancestors = 347587865)

src/performance/memory.ts
sarveshnagpal: Done

---
In reply to: [347588259](https://github.com/microsoft/clarity-js/pull/160#discussion_r347588259) [](ancestors = 347588259)

src/performance/navigation.ts
sarveshnagpal: I was hoping to add these metrics as we have a use for them. Not add everything in first go. What are your thoughts? I've added redirectCount and loadEventStart for completeness. I was in two minds about this. 

---
In reply to: [347587449](https://github.com/microsoft/clarity-js/pull/160#discussion_r347587449) [](ancestors = 347587449)

src/performance/longtask.ts
sarveshnagpal: In practice I wasn't able to see valid values in attribution (self) & (multiple). However, I agree may be should add it and review data at scale. Will add. 

---
In reply to: [347589413](https://github.com/microsoft/clarity-js/pull/160#discussion_r347589413) [](ancestors = 347589413)

src/performance/connection.ts
sarveshnagpal: That's already in performance.d.ts. I had the link and the interface.

src/performance/contentful.ts
sarveshnagpal: Same as above. Already in performance.d.ts where interface are defined.

src/performance/connection.ts
sarveshnagpal: Didn't see it being supported by Chrome. Didn't want to add a null field unless we have a use case for it.

src/layout/dom.ts
sarveshnagpal: Layout was ok when it was all about layout. Now it's more about metadata. I will rename it metadata for now.

src/data/token.ts
sarveshnagpal: Done

decode/performance.ts
sarveshnagpal: Done

src/performance/contentful.ts
toby-walker: Yeah, but that's not a spec, just some TypeScript. For me having the spec link is nice as that's where I go to see what the data should mean and what is "correct". 

---
In reply to: [347636727](https://github.com/microsoft/clarity-js/pull/160#discussion_r347636727) [](ancestors = 347636727)

PR #162, opened 2019-12-04
Bug fixes and improvements in v1.0.0-beta.16

sarveshnagpal:
- Using requestIdleCallback instead of requestAnimationFrame for async tasks

### DIFF
--- a/src/core/task.ts
+++ b/src/core/task.ts
@@ -101,8 +101,7 @@ async function wait(): Promise<RequestIdleCallbackDeadline> {
 //   (2) Use setTimeout call within requestAnimationFrame. This also works, but there's a risk that browser may throttle setTimeout calls.
 // Given this information, we are currently using (1) from above. More information on (2) as well as some additional context is below:
 // https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Performance_best_practices_for_Firefox_fe_engineers
-let requestIdleCallback = window["requestIdleCallback"] ||
-function(callback: (deadline: RequestIdleCallbackDeadline) => void, options: RequestIdleCallbackOptions): void {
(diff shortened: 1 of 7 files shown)


src/core/task.ts
```diff
@@ -49,30 +48,29 @@ function run(): void {
 }
 
 export function shouldYield(method: Metric): boolean {
-    let elapsed = performance.now() - tracker[method];
-    return (elapsed > config.longtask);
+    let elapsed = performance.now() - tracker[method].start;
```
toby-walker: Wouldn't this include time that was not processing? Not sure if your intent is to measure the "real-time" elapsed or the time the task was processing.

src/core/task.ts
```diff
@@ -82,13 +80,46 @@ export async function pause(method: Metric): Promise<void> {
...
+// Background context: requestAnimationFrame invokes the js code right before: style, layout and paint computation within the frame.
+// This means, that any code that runs as part of requestAnimationFrame will by default be blocking in nature. Not what we want.
+// For non-blocking behavior, We need to know when browser has finished painiting. This can be accomplished in two different ways (hacks):
+//   (1) Use MessageChannel to pass the message, and browser will receive the message right after pain event has occured.
+//   (2) Use setTimeout call within requestAnimationFrame. This also works, but there's a risk that browser may throttle setTimeout calls.
+// Given this information, we are currently using (1) from above. More information on (2) as well as some additional context is below:
+// https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Performance_best_practices_for_Firefox_fe_engineers
+let requestIdleCallback = window["requestIdleCallback"] ||
+function(callback: (deadline: RequestIdleCallbackDeadline) => void, options: RequestIdleCallbackOptions): void {
```
toby-walker: Super nit: Could you make this a top-level function so that it is testable and then do window["...."] || myfunction

sarveshnagpal: Done

PR #166, opened 2020-01-18
Bug fixes and improvements in v1.0.0-beta.18

sarveshnagpal:
- Support for pause & resume to mitigate performance conflict with single page applications
- Renaming variables

This is work in progress. Sending out first draft for early review. This is a significant change. I'll follow up later with few more changes, especially around fixing network data bug fixes from b16.

### DIFF
--- a/decode/data.ts
+++ b/decode/data.ts
@@ -1,5 +1,5 @@
 import { BooleanFlag, Envelope, Event, MetricData, PageData, PingData } from "../types/data";
-import { SummaryData, TagData, TargetData, Token, Upload, UploadData } from "../types/data";
+import { SummaryData, TagData, TargetData, Token, UpgradeData, Upload, UploadData } from "../types/data";
 import { DataEvent } from "../types/decode/data";
 
 let summaries: { [key: number]: SummaryData[] } = null;
@@ -35,8 +35,11 @@ export function decode(tokens: Token[]): DataEvent {
                 targetData.push(target);
             }
             return { time, event, data: targetData };
+        case Event.Upgrade:
+            let upgrade: UpgradeData = { key: tokens[2] as string };
+            return { time, event, data: upgrade };
         case Event.Upload:
-            let upload: UploadData = { sequence: tokens[2] as number, attempts: tokens[3] as number, status: tokens[4] as number};
+            let upload: UploadData = { sequence: tokens[2] as number, attempts: tokens[3] as number, status: tokens[4] as number };
             return { time, event, data: upload };
         case Event.Metric:
             let i = 2; // Start from 3rd index since first two are used for time & event
--- a/decode/performance.ts
+++ b/decode/performance.ts
@@ -83,7 +83,11 @@ export function decode(tokens: Token[]): PerformanceEvent  {
                         // Continuing the example used above, following code expands an array of tokens by getting rid of any optimizations
                         // E.g. Current token: [1,3], from array of tokens: ["hello", "world", "coding", "language", [1, 3], "example"]
                         // Following code will expand [1,3] into "world", "language".
-                        if (token === null) { network.push(token); } else {
+                        if (token === null) {
+                            type = null;
+                            if (stringIndex < 3) { stringIndex++; }
+                            network.push(token);
+                        } else {
                             let subtoken = token[0];
                             let subtype = typeof(subtoken);
                             switch (subtype) {
@@ -111,14 +115,13 @@ function process(network: any[] | number[], stringIndex: number): NetworkData {
     // Checking for null values since different browsers implement
     // resource timing spec differently. For instance, not all version of Edge support
     // transferSize property.
-    let empty = "";
     return {
-        start: network[0] === -1 ? null : network[0] as number,
-        duration: network[1] === -1 ? null : network[1] as number,
-        size: network[2] === -1 ? null : network[2] as number,
+        start: network[0] as number,
+        duration: network[1] as number,
+        size: network[2] as number,
         target: stringIndex > 3 ? network[3] as number : null,
-        initiator: network[stringIndex] === empty ? null : network[stringIndex] as string,
-        protocol: network[stringIndex + 1] === empty ? null : network[stringIndex + 1] as string,
-        host: network[stringIndex + 2] === empty ? null : network[stringIndex + 2] as string,
+        initiator: network[stringIndex] as string,
+        protocol: network[stringIndex + 1] as string,
+        host: network[stringIndex + 2] as string,
     };
 }
--- a/decode/performance.ts
+++ b/decode/performance.ts
@@ -108,6 +108,9 @@ export function decode(tokens: Token[]): PerformanceEvent  {
 }
 
 function process(network: any[] | number[], stringIndex: number): NetworkData {
+    // Checking for null values since different browsers implement
+    // resource timing spec differently. For instance, not all version of Edge support
+    // transferSize property.
     let empty = "";
     return {
         start: network[0] === -1 ? null : network[0] as number,
--- a/decode/performance.ts
+++ b/decode/performance.ts
@@ -108,13 +108,14 @@ export function decode(tokens: Token[]): PerformanceEvent  {
 }
 
 function process(network: any[] | number[], stringIndex: number): NetworkData {
+    let empty = "";
     return {
-        start: network[0] as number,
-        duration: network[1] as number,
-        size: network[2] as number,
+        start: network[0] === -1 ? null : network[0] as number,
+        duration: network[1] === -1 ? null : network[1] as number,
+        size: network[2] === -1 ? null : network[2] as number,
         target: stringIndex > 3 ? network[3] as number : null,
-        initiator: network[stringIndex] as string,
-        protocol: network[stringIndex + 1] as string,
-        host: network[stringIndex + 2] as string
+        initiator: network[stringIndex] === empty ? null : network[stringIndex] as string,
+        protocol: network[stringIndex + 1] === empty ? null : network[stringIndex + 1] as string,
+        host: network[stringIndex + 2] === empty ? null : network[stringIndex + 2] as string,
     };
 }
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clarity-js",
-  "version": "1.0.0-beta.17",
+  "version": "1.0.0-beta.18",
   "description": "An analytics library that uses web page interactions to generate aggregated insights",
   "author": "Microsoft Corp.",
   "license": "MIT",
--- a/src/clarity.ts
+++ b/src/clarity.ts
@@ -10,6 +10,7 @@ import * as interaction from "@src/interaction";
 import * as layout from "@src/layout";
 import * as performance from "@src/performance";
 
+const CLARITY = "clarity";
 let status = false;
 
 export function config(override: Config): boolean {
@@ -22,7 +23,9 @@ export function config(override: Config): boolean {
 }
 
 export function start(override: Config = {}): void {
-  if (core.check()) {
+  // Check that browser supports required APIs
+  // And, also that we are not attempting to start Clarity multiple times
+  if (core.check() && status === false) {
     config(override);
     status = true;
 
@@ -37,9 +40,11 @@ export function start(override: Config = {}): void {
 
 function restart(): void {
   start();
+  tag(CLARITY, "restart");
 }
 
 export function suspend(): void {
+  tag(CLARITY, "suspend");
   end();
   bind(document, "mousemove", restart);
   bind(document, "touchstart", restart);
@@ -49,11 +54,13 @@ export function suspend(): void {
 }
 
 export function pause(): void {
+  tag(CLARITY, "pause");
   task.pause();
 }
 
 export function resume(): void {
   task.resume();
+  tag(CLARITY, "resume");
 }
 
 export function end(): void {
--- a/src/clarity.ts
+++ b/src/clarity.ts
@@ -3,6 +3,7 @@ import * as core from "@src/core";
 import configuration from "@src/core/config";
 import { bind } from "@src/core/event";
 import measure from "@src/core/measure";
+import * as task from "@src/core/task";
 import * as data from "@src/data";
 import * as diagnostic from "@src/diagnostic";
 import * as interaction from "@src/interaction";
@@ -34,17 +35,25 @@ export function start(override: Config = {}): void {
   }
 }
 
-export function pause(): void {
+function restart(): void {
+  start();
+}
+
+export function suspend(): void {
   end();
-  bind(document, "mousemove", resume);
-  bind(document, "touchstart", resume);
-  bind(window, "resize", resume);
-  bind(window, "scroll", resume);
-  bind(window, "pageshow", resume);
+  bind(document, "mousemove", restart);
+  bind(document, "touchstart", restart);
+  bind(window, "resize", restart);
+  bind(window, "scroll", restart);
+  bind(window, "pageshow", restart);
+}
+
+export function pause(): void {
+  task.pause();
 }
 
 export function resume(): void {
-  start();
+  task.resume();
 }
 
 export function end(): void {
--- a/src/clarity.ts
+++ b/src/clarity.ts
@@ -76,6 +76,13 @@ export function tag(key: string, value: string): void {
   }
 }
 
+export function upgrade(key: string): void {
+  // Do not process upgrade call if Clarity is not already activated and in lean mode
+  if (status && configuration.lean) {
+    measure(data.upgrade)(key);
+  }
+}
+
 export function active(): boolean {
   return status;
 }
--- a/src/clarity.ts
+++ b/src/clarity.ts
@@ -43,6 +43,14 @@ function restart(): void {
   tag(CLARITY, "restart");
 }
 
+// Suspend ends the current Clarity instance after a configured timeout period
+// The way it differs from the "end" call is that it starts listening to
+// user interaction events as soon as it terminates existing clarity instance.
+// On the next interaction, it automatically starts another instance under a different page id
+// E.g. if configured timeout is 10m, and user stays inactive for an hour.
+// In this case, we will suspend clarity after 10m of inactivity and after another 50m when user interacts again
+// Clarity will restart and start another instance seamlessly. Effectively not missing any active time, but also
+// not holding the session during inactive time periods.
 export function suspend(): void {
   tag(CLARITY, "suspend");
   end();
@@ -53,11 +61,17 @@ export function suspend(): void {
   bind(window, "pageshow", restart);
 }
 
+// By default Clarity is asynchronous and will yield by looking for requestIdleCallback.
+// However, there can still be situations with single page apps where a user action can result
+// in the whole DOM being destroyed and reconstructed. While Clarity will performan favorably out of the box,
+// we do allow external clients to manually pause Clarity for that short burst of time and minimize
+// performance impact even further. For reference, we are talking 10s of milliseconds optimization here, not seconds.
 export function pause(): void {
   tag(CLARITY, "pause");
   task.pause();
 }
 
+// This is how external clients can get out of pause state, and resume Clarity to continue monitoring the page
 export function resume(): void {
   task.resume();
   tag(CLARITY, "resume");
--- a/src/data/index.ts
+++ b/src/data/index.ts
@@ -20,6 +20,9 @@ export function start(): void {
 }
 
 export function end(): void {
+    // The ordering below should respect inter-module dependency.
+    // E.g. if upgrade depends on upload, then upgrade needs to end before upload.
+    // Similarly, if upload depends on metadata, upload needs to end before metadata.
     measure(upgrade.reset)();
     measure(tag.reset)();
     measure(ping.end)();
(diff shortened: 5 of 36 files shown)


decode/data.ts
```diff
@@ -35,8 +35,11 @@ export function decode(tokens: Token[]): DataEvent {
...
             }
             return { time, event, data: targetData };
+        case Event.Upgrade:
+            let upgrade: UpgradeData = { key: tokens[2] as string };
+            return { time, event, data: upgrade };
         case Event.Upload:
-            let upload: UploadData = { sequence: tokens[2] as number, attempts: tokens[3] as number, status: tokens[4] as number};
+            let upload: UploadData = { sequence: tokens[2] as number, attempts: tokens[3] as number, status: tokens[4] as number };
             return { time, event, data: upload };
         case Event.Metric:
```
toby-walker, line 44: Add a default clause here in case you miss something. Similarly for all the big case of event clauses in the program (check them all). #Pending

decode/performance.ts
```diff
@@ -108,13 +108,14 @@ export function decode(tokens: Token[]): PerformanceEvent  {
 }
 
 function process(network: any[] | number[], stringIndex: number): NetworkData {
+    let empty = "";
     return {
-        start: network[0] as number,
-        duration: network[1] as number,
-        size: network[2] as number,
+        start: network[0] === -1 ? null : network[0] as number,
```
toby-walker: When would this be null? Similar for the other properties here. #Resolved

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clarity-js",
-  "version": "1.0.0-beta.17",
```
toby-walker: What was in b17? We never shipped it in a flight did we?

src/clarity.ts
```diff
@@ -34,17 +35,25 @@ export function start(override: Config = {}): void {
   }
 }
 
-export function pause(): void {
+function restart(): void {
+  start();
+}
+
+export function suspend(): void {
```
toby-walker, line 54: Do we send specific events when we go through these state transitions (restart/pause/suspend)? #Resolved

src/clarity.ts
```diff
@@ -34,17 +35,25 @@ export function start(override: Config = {}): void {
...
   end();
-  bind(document, "mousemove", resume);
-  bind(document, "touchstart", resume);
-  bind(window, "resize", resume);
-  bind(window, "scroll", resume);
-  bind(window, "pageshow", resume);
+  bind(document, "mousemove", restart);
+  bind(document, "touchstart", restart);
+  bind(window, "resize", restart);
```
toby-walker, line 59: Will multiple calls to restart be handled correctly (potentially a race condition if there are multiple events queued up. Almost surely would happen with mousemove (mulitple events triggered). #Closed

toby-walker:
> export interface MetricEvent extends PartialEvent { data: MetricData; }

Aside: There's a standard way for declaring this kind of thing using templates. 

PageEvent extends PartialEvent<PageData>

Where PartialEvent<T> is defined as 
interface<T> {
data:T
}

Not sure about the exact TypeScript syntax.  #Pending

---
Refers to: types/decode/data.d.ts:5 in fc920a3. [](commit_id = fc920a336e55554fa7250cf07c6141d9c2bc8b91, deletion_comment = False)

types/data.d.ts
```diff
@@ -43,7 +43,9 @@ export const enum Event {
     Paint = 35,
     ContentfulPaint = 36,
     Memory = 37,
-    Connection = 38
+    Connection = 38,
+    Backup = 39,
```
toby-walker: Briefly what are the semantics of these 2 new events (backup and upgrade)?

src/performance/network.ts
```diff
@@ -18,13 +19,13 @@ export function compute(entry: PerformanceResourceTiming): void {
     state.push({
         url: entry.name,
         data: {
-            start: Math.round(entry.startTime),
-            duration: Math.round(entry.duration),
-            size: entry.transferSize,
+            start: "startTime" in entry ? Math.round(entry.startTime) : -1,
```
toby-walker: Why don't you just push null here and then not have the extra complexity in other file? You go back and forth between -1/null (for numeric) and empty/null for strings.  #Resolved

src/performance/encode.ts
```diff
@@ -73,13 +73,13 @@ export default async function(type: Event): Promise<void> {
...
-            queue(tokens);
+            queue(tokens, type);
             break;
         case Event.Network:
             for (let state of network.state) {
-                if (task.shouldYield(timer)) { await task.pause(timer); }
+                if (task.shouldYield(timer)) { await task.suspend(timer); }
                 let data = state.data;
-                data.target = observe(getMatch(state.url));
+                data.target = observe({id: null, selector: null, node: getMatch(state.url)});
```
toby-walker: >observe [](start = 30, length = 7)

Why are id and selector null here? #Closed

src/layout/encode.ts
```diff
@@ -66,7 +67,8 @@ export default async function(type: Event): Promise<void> {
                 }
                 tokens = tokenize(tokens, metadata);
             }
-            queue(tokens);
+
+            queue(tokens, config.lean ? Event.Backup : type);
```
toby-walker: What is this doing and what is Event.Backup? #Closed

src/interaction/selection.ts
```diff
@@ -27,15 +28,16 @@ function recompute(): void {
     // can result in null anchorNode and focusNode if there was no previous selection on page
     if (current.anchorNode === null && current.focusNode === null) { return; }
 
-    if (previous !== null && data.start !== null && data.start !== current.anchorNode) {
+    let startNode = data.start ? (data.start as TargetInfo).node : null;
```
toby-walker, line 31: >startNode [](start = 8, length = 9)

Knowing the start and end elements isn't really that useful. Can you send the complete list of selected elements instead? #Pending

src/diagnostic/encode.ts
```diff
@@ -15,13 +15,13 @@ export default async function(type: Event): Promise<void> {
             tokens.push(script.data.column);
             tokens.push(script.data.stack);
             tokens.push(script.data.source);
-            queue(tokens);
+            queue(tokens, type);
             break;
         case Event.ImageError:
             if (image.data) {
```
toby-walker, line 21: When would this if condition not hold?

src/data/index.ts
```diff
@@ -4,8 +4,10 @@ import * as metric from "@src/data/metric";
 import * as ping from "@src/data/ping";
 import * as tag from "@src/data/tag";
 import * as target from "@src/data/target";
+import * as upgrade from "@src/data/upgrade";
 import * as upload from "@src/data/upload";
 export { tag } from "@src/data/tag";
+export { upgrade } from "@src/data/upgrade";
 
 export function start(): void {
```
toby-walker, line 12: What determines which of reset/start/end is used in thestart() and end() functions below? Wasn't quite sure of the logic.  #Pending

src/data/index.ts
```diff
@@ -14,9 +16,11 @@ export function start(): void {
     measure(metadata.start)();
     measure(ping.start)();
     measure(tag.reset)();
+    measure(upgrade.reset)();
 }
 
 export function end(): void {
```
toby-walker, line 22: Is the reverse ordering of calls in start and end important? If so, add a comment about why.  #Closed

src/data/upload.ts
```diff
@@ -19,29 +20,35 @@ export let track: UploadData;
...
 }
 
-export function queue(data: Token[]): void {
+export function queue(data: Token[], type: Event): void {
     if (active) {
-        let type = data.length > 1 ? data[1] : null;
         let event = JSON.stringify(data);
-        events.push(event);
+        let transmit = true; // schedule upload callback
+        let container = events;
```
toby-walker, line 36: >container [](start = 12, length = 9)

Why are these events transmit = false? Add comment. #Closed

src/data/upload.ts
```diff
@@ -57,17 +64,29 @@ export function queue(data: Token[]): void {
             case Event.ScriptError:
             case Event.ImageError:
                 break;
+            case Event.Backup:
+                transmit = false;
+                container = backup;
+                break;
+            case Event.Upgrade:
+                for (let entry of backup) {
```
toby-walker, line 87: What is this logic doing? #Closed

src/data/upgrade.ts
```diff
@@ -0,0 +1,15 @@
...
+import config from "@src/core/config";
+import encode from "@src/data/encode";
+
+export let data: UpgradeData = null;
+
+export function reset(): void {
+    data = null;
+}
+
+export function upgrade(key: string): void {
```
toby-walker, line 15: >upgrade [](start = 16, length = 7)

Who calls this function and when? What do we do when we upgrade? A bit of a higher level design comment would be helpful.  #Closed

toby-walker (changes requested):
:clock1:

decode/data.ts
sarveshnagpal: It's a good point. I'll cover data quality changes in another round. If I add default, I will need to figure out what to do with these unhandled cases. Probably throwing an exception is ok but worth thinking about more.

---
In reply to: [368662595](https://github.com/microsoft/clarity-js/pull/166#discussion_r368662595) [](ancestors = 368662595)

decode/performance.ts
sarveshnagpal: Previous issues that you reported were due to the fact that different browsers implement spec differently. For now, the known field is "transferSize". For instance, not all versions of Edge support it. I've added the check to see if we get null across any browser and if we don't, we can remove the future check. But today if we get null, the decoding will go wrong and it will yield random output like you shared in the email. 

---
In reply to: [368662863](https://github.com/microsoft/clarity-js/pull/166#discussion_r368662863) [](ancestors = 368662863)

sarveshnagpal: I've also added comment.

---
In reply to: [368802956](https://github.com/microsoft/clarity-js/pull/166#discussion_r368802956) [](ancestors = 368802956,368662863)

src/clarity.ts
sarveshnagpal: Good point. For suspend yes. However, for pause & resume we don't. For now I will use tags to add instrumentation for these actions so we know when these actions happened.

---
In reply to: [368664746](https://github.com/microsoft/clarity-js/pull/166#discussion_r368664746) [](ancestors = 368664746)

src/clarity.ts
sarveshnagpal: This is not new logic so if it was handled before, it will continue to be the same deal. However going through the code we were unnecessarily re-initializing modules twice on multiple start calls. I've added the check to prevent that from happening. Thanks for pointing it out.

---
In reply to: [368665036](https://github.com/microsoft/clarity-js/pull/166#discussion_r368665036) [](ancestors = 368665036)

types/data.d.ts
sarveshnagpal: Backup => It backs up any layout events in lean mode until clarity.upgrade() is called. So instead of throwing away layout events like before, now we store them in memory. 

Upgrade => This event retrieves stored events in backup and sends them to server. When this happens, lean mode  (boxmodel like) page seamless upgrades into full blow session from the beginning of this page start time.

---
In reply to: [368666638](https://github.com/microsoft/clarity-js/pull/166#discussion_r368666638) [](ancestors = 368666638)

src/performance/encode.ts
sarveshnagpal: Simplified this to match rest of the coding pattern. No more setting null values. 

---
In reply to: [368667819](https://github.com/microsoft/clarity-js/pull/166#discussion_r368667819) [](ancestors = 368667819)

src/layout/encode.ts
sarveshnagpal: Added comments to clarify. I also explained Event.Backup in a previous comment; it's to back up layout events in memory.

---
In reply to: [368677998](https://github.com/microsoft/clarity-js/pull/166#discussion_r368677998) [](ancestors = 368677998)

src/interaction/selection.ts
sarveshnagpal: Leaving this out of scope for now. This is on my list but it will be a slightly bigger change. Will get to it in next version.

---
In reply to: [368678337](https://github.com/microsoft/clarity-js/pull/166#discussion_r368678337) [](ancestors = 368678337)

src/diagnostic/encode.ts
sarveshnagpal: We listen to high level error handler. Browser will fire this event for non image elements too. e.g. canvas. For now I'm restricting to image because I'm not sure what would be a good way to handle other elements. In future we can consider expanding this. At the moment it's focused on image failures.

---
In reply to: [368679267](https://github.com/microsoft/clarity-js/pull/166#discussion_r368679267) [](ancestors = 368679267)

src/data/index.ts
sarveshnagpal: I can walk you through it offline. Anything that just resets the state, the module exposes "reset" function. Anything that needs to initialize some variables and then later clear state, those modules use start / end. Should we later unify for consistency? Probably. 

---
In reply to: [368679796](https://github.com/microsoft/clarity-js/pull/166#discussion_r368679796) [](ancestors = 368679796)

src/data/index.ts
sarveshnagpal: Not entirely. Order can be mixed up for independent modules however for modules with internal dependency - e.g. -  upload takes dependency on metadata, we can't end metadata before upload or there will be script errors. I will add this comment. 

---
In reply to: [368679924](https://github.com/microsoft/clarity-js/pull/166#discussion_r368679924) [](ancestors = 368679924)

src/data/upload.ts
sarveshnagpal: Done

---
In reply to: [368680984](https://github.com/microsoft/clarity-js/pull/166#discussion_r368680984) [](ancestors = 368680984)

src/data/upload.ts
sarveshnagpal: Explained in comments. It looks up backed up events and gets them ready to be sent to the server. This allows transition from lean mode into full mode.

---
In reply to: [368681147](https://github.com/microsoft/clarity-js/pull/166#discussion_r368681147) [](ancestors = 368681147)

src/data/upgrade.ts
sarveshnagpal: Good point, done.

---
In reply to: [368681354](https://github.com/microsoft/clarity-js/pull/166#discussion_r368681354) [](ancestors = 368681354)

sarveshnagpal:
> export interface MetricEvent extends PartialEvent { data: MetricData; }

Thanks, I will look it up but probably as part of the next change.

---
In reply to: [576374275](https://github.com/microsoft/clarity-js/pull/166#issuecomment-576374275) [](ancestors = 576374275)

---
Refers to: types/decode/data.d.ts:5 in fc920a3. [](commit_id = fc920a336e55554fa7250cf07c6141d9c2bc8b91, deletion_comment = False)

src/performance/network.ts
sarveshnagpal: Thinking more on this. Leaving this comment open.

---
In reply to: [368666995](https://github.com/microsoft/clarity-js/pull/166#discussion_r368666995) [](ancestors = 368666995)

decode/performance.ts
toby-walker: Why does decoding go wrong with null values? Null seems like a perfectly legitimate values that the encoding/decoding should handle. 

---
In reply to: [368802993](https://github.com/microsoft/clarity-js/pull/166#discussion_r368802993) [](ancestors = 368802993,368802956,368662863)

decode/performance.ts
```diff
@@ -108,13 +108,17 @@ export function decode(tokens: Token[]): PerformanceEvent  {
...
+        duration: network[1] === -1 ? null : network[1] as number,
+        size: network[2] === -1 ? null : network[2] as number,
         target: stringIndex > 3 ? network[3] as number : null,
-        initiator: network[stringIndex] as string,
-        protocol: network[stringIndex + 1] as string,
-        host: network[stringIndex + 2] as string
```
toby-walker, line 118: Please fix the typing declaration ni thing like NetworkData to indicate that the field is nullable. For example it's like this now:

export interface NetworkData {
    start: number;
    duration: number;
    size: number;
    target: Target;
    initiator: string;
    protocol: string;
    host: string;
}

I think it should be something like: 

export interface NetworkData {
    start: number | null;
    duration: number | null;
    size: number | null
    target: Target;
    initiator?: string;
    protocol?: string;
    host?: string;
}

One other nuance is the difference between attribute not present vs. attribute is present and null. See above where I wasn't sure in some cases. Please check other cases where you have this kind if "null handling" for variably-supported features. 

And that's a best guess It would be very helpful to get this precise as it's what I use to define the matching schema on the pipeline were what is optional, or nullable (not same) needs to be exactly in sync with your code. 

 #WontFix

src/clarity.ts
toby-walker: Could you add a comment here explaining this a bit?

---
In reply to: [368803710](https://github.com/microsoft/clarity-js/pull/166#discussion_r368803710) [](ancestors = 368803710,368664746)

types/data.d.ts
toby-walker: Add comment please with above. This is as good a place as any for the comment. 

Reading above I'm still not clear:

* What exactly does a backup event represent and is it internal or is something sent with data? If so, when is the event sent and what data does it contain? I assume it doesn't contain
* Sam question for the upgrade event. What you are describing above is the behavior of the client code, not the event the data semantics.  Events don't *do* anything they represent something happening. 
* It might be good to bundle these "client state transition" events into a common event as each kind of event ends up having overhead to add. Judgment call. 

---
In reply to: [368805833](https://github.com/microsoft/clarity-js/pull/166#discussion_r368805833) [](ancestors = 368805833,368666638)

src/data/upgrade.ts
```diff
@@ -0,0 +1,19 @@
...
+import config from "@src/core/config";
+import encode from "@src/data/encode";
+
+export let data: UpgradeData = null;
+
+export function reset(): void {
+    data = null;
+}
+
+// Following call will upgrade the session from lean mode into the full mode retroactively from the start of the page.
```
toby-walker, line 11: Who calls this function and when? It is supposed to be externally triggered?  #Resolved

src/diagnostic/encode.ts
toby-walker: Add above as comment. 

---
In reply to: [368807461](https://github.com/microsoft/clarity-js/pull/166#discussion_r368807461) [](ancestors = 368807461,368679267)

src/layout/encode.ts
```diff
@@ -66,7 +67,12 @@ export default async function(type: Event): Promise<void> {
                 }
                 tokens = tokenize(tokens, metadata);
             }
-            queue(tokens);
+
+            // Layout events are queued based on the active configuration
+            // If lean mode is on, instead of sending these events to server, we back them up in memory using Event.Backup.
```
toby-walker: Will we limit the amount of data queued up for this? Seems like it could grow too large if we don't cap it. Maybe cap at a max size of X KB.

src/performance/network.ts
sarveshnagpal: Simplified the code and using null for null, and only for fields where it makes sense.

---
In reply to: [368810644](https://github.com/microsoft/clarity-js/pull/166#discussion_r368810644) [](ancestors = 368810644,368666995)

decode/performance.ts
sarveshnagpal: Simplified this whole code. Not using null anymore. 

---
In reply to: [369054487](https://github.com/microsoft/clarity-js/pull/166#discussion_r369054487) [](ancestors = 369054487,368802993,368802956,368662863)

decode/performance.ts
sarveshnagpal: I've simplified this. (?) in typescript means the field is optional, which these are not. However if the value is missing, it will acquire a null value - which is a valid type: number or string. This is consistent with other places in the code. Take a look with refactored code and let me know your thoughts. Resolving this for now. 

---
In reply to: [369057739](https://github.com/microsoft/clarity-js/pull/166#discussion_r369057739) [](ancestors = 369057739)

src/data/upgrade.ts
sarveshnagpal: Yes, you will notice the call in SNR code flow that I will send out shortly. Similar to start function too - it's called externally. 

---
In reply to: [369061624](https://github.com/microsoft/clarity-js/pull/166#discussion_r369061624) [](ancestors = 369061624)

src/layout/encode.ts
sarveshnagpal: This is tricky; let's talk later today about this. It was a concern in my mind too but if we pause this data, or do FIFO - the data quality won't hold anymore. Leaving this open for us to discuss later. 

---
In reply to: [369063178](https://github.com/microsoft/clarity-js/pull/166#discussion_r369063178) [](ancestors = 369063178)

src/clarity.ts
sarveshnagpal: Done, thanks.

---
In reply to: [369058546](https://github.com/microsoft/clarity-js/pull/166#discussion_r369058546) [](ancestors = 369058546,368803710,368664746)

toby-walker (approved):

PR #170, opened 2020-03-18
Bug fixes and improvements in v1.0.0-beta.20

sarveshnagpal:
- Improving session tracking and support for cookie-less instrumentation
- Support for regions
- Better lean mode visualization

### DIFF
--- a/decode/clarity.ts
+++ b/decode/clarity.ts
@@ -227,10 +227,6 @@ export async function replay(
                 let boxModelEvent = entry as BoxModelEvent;
                 r.boxmodel(boxModelEvent.data, iframe);
                 break;
-            case Event.Target:
-                let targetEvent = entry as TargetEvent;
-                r.target(targetEvent.data, iframe);
-                break;
             case Event.MouseDown:
             case Event.MouseUp:
             case Event.MouseMove:
--- a/decode/data.ts
+++ b/decode/data.ts
@@ -30,8 +30,13 @@ export function decode(tokens: Token[]): DataEvent {
             return { time, event, data: tag };
         case Event.Target:
             let targetData: TargetData[] = [];
-            for (let t = 2; t < tokens.length; t += 3) {
-                let target: TargetData = { id: tokens[t] as number, hash: tokens[t + 1] as string, box: tokens[t + 2] as number[] };
+            for (let t = 2; t < tokens.length; t += 4) {
+                let target: TargetData = {
+                    id: tokens[t] as number,
+                    hash: tokens[t + 1] as string,
+                    region: tokens[t + 2] as string,
+                    box: tokens[t + 3] as number[]
+                };
                 targetData.push(target);
             }
             return { time, event, data: targetData };
--- a/decode/layout.ts
+++ b/decode/layout.ts
@@ -25,8 +25,8 @@ export function decode(tokens: Token[]): LayoutEvent {
             return { time, event, data: documentData };
         case Event.BoxModel:
             let boxmodelData: BoxModelData[] = [];
-            for (let i = 2; i < tokens.length; i += 2) {
-                let boxmodel: BoxModelData = { id: tokens[i] as number, box: tokens[i + 1] as number[] };
+            for (let i = 2; i < tokens.length; i += 3) {
+                let boxmodel: BoxModelData = { id: tokens[i] as number, region: tokens[i + 1] as string, box: tokens[i + 2] as number[] };
                 boxmodelData.push(boxmodel);
             }
             return { time, event, data: boxmodelData };
--- a/decode/layout.ts
+++ b/decode/layout.ts
@@ -26,7 +26,7 @@ export function decode(tokens: Token[]): LayoutEvent {
         case Event.BoxModel:
             let boxmodelData: BoxModelData[] = [];
             for (let i = 2; i < tokens.length; i += 3) {
-                let boxmodel: BoxModelData = { id: tokens[i] as number, region: tokens[i + 1] as string, box: tokens[i + 2] as number[] };
+                let boxmodel: BoxModelData = { id: tokens[i] as number, box: tokens[i + 1] as number[], region: tokens[i + 2] as string };
                 boxmodelData.push(boxmodel);
             }
             return { time, event, data: boxmodelData };
--- a/src/clarity.ts
+++ b/src/clarity.ts
@@ -104,7 +104,16 @@ export function tag(key: string, value: string): void {
 export function upgrade(key: string): void {
   // Do not process upgrade call if Clarity is not already activated and in lean mode
   if (status && configuration.lean) {
+    configuration.lean = false;
     measure(data.upgrade)(key);
+  }
+}
+
+export function consent(): void {
+  // Do not begin tracking user if Clarity is not already activated
+  if (status && !configuration.track) {
+    configuration.track = true;
+    measure(data.track)();
   }
 }
 
--- a/src/clarity.ts
+++ b/src/clarity.ts
@@ -104,7 +104,6 @@ export function tag(key: string, value: string): void {
 export function upgrade(key: string): void {
   // Do not process upgrade call if Clarity is not already activated and in lean mode
   if (status && configuration.lean) {
-    configuration.lean = false;
     measure(data.upgrade)(key);
   }
 }
@@ -112,8 +111,7 @@ export function upgrade(key: string): void {
 export function consent(): void {
   // Do not begin tracking user if Clarity is not already activated
   if (status && !configuration.track) {
-    configuration.track = true;
-    measure(data.track)();
+    measure(data.consent)();
   }
 }
 
--- a/src/core/config.ts
+++ b/src/core/config.ts
@@ -1,4 +1,4 @@
-import { Config } from "@clarity-types/core";
+import { Config, Time } from "@clarity-types/core";
 
 let config: Config = {
     projectId: null,
@@ -6,12 +6,12 @@ let config: Config = {
     lookahead: 500, // 500 milliseconds
     distance: 20, // 20 pixels
     interval: 25, // 25 milliseconds
-    delay: 1000, // 1 second
+    delay: 1 * Time.Second, // 1 second
     expire: 365, // 1 year
     ping: 60 * 1000, // 1 minute
-    timeout: 5 * 60 * 1000, // 5 minutes
-    session: 10 * 60 * 1000, // 5 minutes
-    shutdown: 2 * 60 * 60 * 1000, // 2 hours
+    timeout: 5 * Time.Minute, // 5 minutes
+    session: 10 * Time.Minute, // 10 minutes
+    shutdown: 2 * Time.Hour, // 2 hours
     cssRules: false,
     lean: false,
     track: true,
--- a/src/core/config.ts
+++ b/src/core/config.ts
@@ -7,12 +7,14 @@ let config: Config = {
     distance: 20, // 20 pixels
     interval: 25, // 25 milliseconds
     delay: 1000, // 1 second
-    expire: 7, // 7 days
+    expire: 365, // 1 year
     ping: 60 * 1000, // 1 minute
     timeout: 5 * 60 * 1000, // 5 minutes
+    session: 10 * 60 * 1000, // 5 minutes
     shutdown: 2 * 60 * 60 * 1000, // 2 hours
     cssRules: false,
     lean: false,
+    track: true,
     tokens: [],
     url: "",
     onstart: null,
--- a/src/core/config.ts
+++ b/src/core/config.ts
@@ -15,7 +15,7 @@ let config: Config = {
     cssRules: false,
     lean: false,
     track: true,
-    tokens: [],
+    regions: {},
     url: "",
     onstart: null,
     upload: null
--- a/src/data/metadata.ts
+++ b/src/data/metadata.ts
@@ -40,7 +40,12 @@ export function envelope(last: boolean): Token[] {
     return [e.sequence, e.version, e.projectId, e.userId, e.sessionId, e.pageId, e.upload, e.end];
 }
 
-export function track(): void {
+export function consent(): void {
+  config.track = true;
+  track();
+}
+
+function track(): void {
   if (config.track) {
     let expiry = new Date();
     expiry.setDate(expiry.getDate() + config.expire);
--- a/src/data/metadata.ts
+++ b/src/data/metadata.ts
@@ -76,7 +76,7 @@ function session(ts: number): string {
       let parts = value.split(CLARITY_STORAGE_SEPARATOR);
       if (parts.length === 2 && ts - parseInt(parts[1], 10) < config.session) { id = parts[0]; }
     }
-    sessionStorage.setItem(CLARITY_STORAGE_KEY, `${id}|${ts}`);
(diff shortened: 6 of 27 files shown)


decode/clarity.ts
```diff
@@ -227,10 +227,6 @@ export async function replay(
                 let boxModelEvent = entry as BoxModelEvent;
                 r.boxmodel(boxModelEvent.data, iframe);
                 break;
-            case Event.Target:
```
toby-walker, line 230: Are you removing this? That would break things.  #Closed

decode/data.ts
```diff
@@ -30,8 +30,13 @@ export function decode(tokens: Token[]): DataEvent {
             return { time, event, data: tag };
         case Event.Target:
             let targetData: TargetData[] = [];
-            for (let t = 2; t < tokens.length; t += 3) {
-                let target: TargetData = { id: tokens[t] as number, hash: tokens[t + 1] as string, box: tokens[t + 2] as number[] };
+            for (let t = 2; t < tokens.length; t += 4) {
+                let target: TargetData = {
+                    id: tokens[t] as number,
+                    hash: tokens[t + 1] as string,
+                    region: tokens[t + 2] as string,
```
toby-walker, line 37: >region [](start = 20, length = 6)

What does this represent? #Closed

decode/layout.ts
```diff
@@ -25,8 +25,8 @@ export function decode(tokens: Token[]): LayoutEvent {
             return { time, event, data: documentData };
         case Event.BoxModel:
             let boxmodelData: BoxModelData[] = [];
-            for (let i = 2; i < tokens.length; i += 2) {
-                let boxmodel: BoxModelData = { id: tokens[i] as number, box: tokens[i + 1] as number[] };
+            for (let i = 2; i < tokens.length; i += 3) {
+                let boxmodel: BoxModelData = { id: tokens[i] as number, region: tokens[i + 1] as string, box: tokens[i + 2] as number[] };
```
toby-walker: Seems better to add this at end as it would be more compatible.  #Closed

src/clarity.ts
```diff
@@ -104,10 +104,19 @@ export function tag(key: string, value: string): void {
...
+    configuration.lean = false;
     measure(data.upgrade)(key);
   }
 }
 
+export function consent(): void {
+  // Do not begin tracking user if Clarity is not already activated
+  if (status && !configuration.track) {
+    configuration.track = true;
+    measure(data.track)();
```
toby-walker: >track [](start = 17, length = 5)

Could we call this consent to make it clearer? Track is vague.  #Closed

src/core/config.ts
```diff
@@ -7,13 +7,15 @@ let config: Config = {
     distance: 20, // 20 pixels
     interval: 25, // 25 milliseconds
     delay: 1000, // 1 second
-    expire: 7, // 7 days
+    expire: 365, // 1 year
     ping: 60 * 1000, // 1 minute
     timeout: 5 * 60 * 1000, // 5 minutes
+    session: 10 * 60 * 1000, // 5 minutes
```
toby-walker: Minor suggestion: Put some constant like Time.SecondMs, Time.MinuteMs, Time.HourMs to make these easier to read. #Closed

types/core.d.ts
```diff
@@ -16,6 +16,10 @@ export interface TaskTracker {
     [key: number]: TaskInfo;
 }
 
+export interface Region {
```
toby-walker: How does this relate to properties called "region" in a few places which are string? It would be great to use a clearer or more qualified name here. 
My suggestion: RegionConfig or TrackingRegions or similar.  #Closed

types/core.d.ts
```diff
@@ -16,6 +16,10 @@ export interface TaskTracker {
...
+    [key: string]: string;
```
toby-walker, line 27: What do the key and string value represent? Is the value a CSS selector? #Closed

types/core.d.ts
```diff
@@ -53,10 +57,12 @@ export interface Config {
     expire?: number;
     ping?: number;
     timeout?: number;
+    session?: number;
     shutdown?: number;
     cssRules?: boolean;
     lean?: boolean;
-    tokens?: string[];
+    track?: boolean;
```
toby-walker, line 71: What does this mean? #Closed

src/layout/mutation.ts
```diff
@@ -74,17 +75,20 @@ async function process(): Promise<void> {
...
             if (task.shouldYield(timer)) { await task.suspend(timer); }
+            dom.regions(target as HTMLElement);
             processNode(target, Source.CharacterData);
             break;
         case "childList":
           // Process additions
           let addedLength = mutation.addedNodes.length;
           for (let j = 0; j < addedLength; j++) {
             let addedNode = mutation.addedNodes[j];
+            dom.regions(addedNode as HTMLElement);
```
toby-walker: >regions [](start = 16, length = 7)

What does this do? Make this a verb, not a noun. #Closed

src/layout/dom.ts
```diff
@@ -21,9 +24,24 @@ export function reset(): void {
     selectorMap = [];
     urlMap = {};
     idMap = new WeakMap();
+    regionMap = new WeakMap();
     if (Constant.DEVTOOLS_HOOK in window) { window[Constant.DEVTOOLS_HOOK] = { get, getNode, history }; }
 }
 
+export function regions(root: ParentNode): void {
+    for (let key in config.regions) {
+        if (config.regions[key] && "querySelectorAll" in root) {
```
toby-walker, line 39: If this doesn't exist is there a way for us to know and check earlier rather than here? If it doesn't exist should we even run Clarity? My prefernce is that we check once at init time and not enable Clarity if its feature prereqs fail. #Closed

src/layout/dom.ts
```diff
@@ -21,9 +24,24 @@ export function reset(): void {
...
+            let elements = root.querySelectorAll(config.regions[key]);
+            let length = elements.length;
+            for (let i = 0; i < length; i++) {
+                if (!(key in regionTracker)) { regionTracker[key] = 0; }
```
toby-walker, line 43: >regionTracker [](start = 29, length = 13)

Can you add a comment explaining what this line with regionMap is doing? Why is a weak map? #Closed

src/layout/dom.ts
```diff
@@ -41,15 +59,20 @@ export function add(node: Node, data: NodeInfo, source: Source): void {
     let nextId = getNextId(node);
     let masked = true;
     let parent = null;
+    let region = regionMap.has(node) ? regionMap.get(node) : null;
 
     if (parentId >= 0 && values[parentId]) {
         parent = values[parentId];
         parent.children.push(id);
+        region = region === null ? parent.region : region;
```
toby-walker, line 73: >region [](start = 42, length = 6)

is this guarantee to exist or do you need to check? #Closed

src/layout/dom.ts
```diff
@@ -41,15 +59,20 @@ export function add(node: Node, data: NodeInfo, source: Source): void {
...
         masked = parent.metadata.masked;
     }
 
     if (data.attributes && Constant.MASK_ATTRIBUTE in data.attributes) { masked = true; }
     if (data.attributes && Constant.UNMASK_ATTRIBUTE in data.attributes) { masked = false; }
+    if (data.attributes && Constant.CLARITY_ID_ATTRIBUTE in data.attributes) {
+        regionMap.set(node, data.attributes[Constant.CLARITY_ID_ATTRIBUTE]);
```
toby-walker: >CLARITY_ID_ATTRIBUTE [](start = 53, length = 20)

What is this and it's not thee "id" attribute, right? #Closed

src/layout/dom.ts
```diff
@@ -270,13 +280,12 @@ function metadata(tag: string, id: number, parentId: number): void {
...
-                break;
-            default:
-                // Capture layout for any element with a user defined selector
-                value.metadata.boxmodel = value.selector.indexOf("*") === 0;
+                if (config.lean === false) { value.metadata.boxmodel = true; }
                 break;
         }
+
+        // Enable boxmodel if this node defines a new region
+        if (regionMap.has(nodes[id])) { value.metadata.boxmodel = true; }
```
toby-walker, line 297: >regionMap [](start = 12, length = 9)

Does this mean we'll track box model only for specified nodes?
Also is the tracking recursive so we track the child if the parent is tracked? #Closed

src/data/metadata.ts
```diff
@@ -54,32 +62,39 @@ function guid() {
...
-  document.cookie = CLARITY_COOKIE_NAME + "=" + value;
+function session(ts: number): string {
+  let id = str(ts, 36);
+  if (config.track && sessionStorage) {
+    let value = sessionStorage.getItem(CLARITY_STORAGE_KEY);
+    if (value && value.indexOf(CLARITY_STORAGE_SEPARATOR) >= 0) {
+      let parts = value.split(CLARITY_STORAGE_SEPARATOR);
+      if (parts.length === 2 && ts - parseInt(parts[1], 10) < config.session) { id = parts[0]; }
+    }
+    sessionStorage.setItem(CLARITY_STORAGE_KEY, `${id}|${ts}`);
```
toby-walker: >} [](start = 53, length = 1)

Should use the STORAGE_SEPARATOR here. It's hardcoded in the string.  #Closed

src/data/metadata.ts
```diff
@@ -54,32 +62,39 @@ function guid() {
...
   let cookies: string[] = document.cookie.split(";");
   if (cookies) {
     for (let i = 0; i < cookies.length; i++) {
       let pair: string[] = cookies[i].split("=");
-      if (pair.length > 1 && pair[0].indexOf(CLARITY_COOKIE_NAME) >= 0 && pair[1].indexOf(CLARITY_COOKIE_SEPARATOR) > 0) {
-        let parts = pair[1].split(CLARITY_COOKIE_SEPARATOR);
-        if (parts.length === 3) {
-          return { userId: parts[0], sessionId: parts[1], timestamp: parseInt(parts[2], 10) };
-        }
+      if (pair.length > 1 && pair[0].indexOf(CLARITY_STORAGE_KEY) >= 0 && pair[1].length === 32) {
```
toby-walker, line 99: >CLARITY_STORAGE_KEY [](start = 45, length = 19)

Shouldn't it generate the guid() only when it fails to retrieve it? Optimization only -- looks correct as written.  #Closed

toby-walker (dismissed):
:clock1:

decode/clarity.ts
sarveshnagpal: No, this is only for rendering Target events in visualization of lean mode. This will have no impact to the data you receive, we are still parsing Event.Target like before.  #Closed

decode/data.ts
sarveshnagpal: A region on the page that may include multiple elements. This is similar to box model selector concept. #Closed

src/clarity.ts
sarveshnagpal: Sure, I will implement the change. #Closed

decode/layout.ts
sarveshnagpal: Sounds good, I will do it and then resolve the comment. #Closed

src/core/config.ts
sarveshnagpal: Sounds good, will do and then resolve. #Closed

types/core.d.ts
sarveshnagpal: Sounds good, will change and then resolve. I will think of a name. #Closed

types/core.d.ts
sarveshnagpal: Yes; I can add comment to explain. #Closed

types/core.d.ts
sarveshnagpal: This was a dead reference in the code for some time. This is left over from our previous discussions on ambition client side compression by sending tokens from the server. We are not going with it right now... and all related code is already out, so I also removed these references for now. In short, think of it as dead code. Will not impact backend. #Closed

src/layout/mutation.ts
sarveshnagpal: Yes, I will make it findRegions and will resolve. #Closed

src/layout/dom.ts
sarveshnagpal: This is mainly for nodes that can get mutation but do not support querySelectorAll. E.g. text nodes. We need to proactively look for regions as single page app modifies itself. In those cases we want to ensure that we ignore mutations on nodes like text, comment, etc. I will think through this a bit more. #Closed

src/layout/dom.ts
sarveshnagpal: It is data-clarity tag. I will change it to data-clarity-region to be consistent with rest of the code. I will make the change. #Closed

src/layout/dom.ts
sarveshnagpal: Yes, only specific nodes. No, it is not recursive and the key point here is that we don't have to track the child. It will capture the region and that's it. #Closed

src/data/metadata.ts
sarveshnagpal: Thanks, will fix. #Closed

src/clarity.ts
sarveshnagpal: On second thoughts after making the change I think the previous version was better. Because we do expose consent call to the user of the API. Once it's called, we update the config and then  start tracking the user. If I renamed the function to consent, it wouldn't make much sense in the default case where we use it to track the user. #Closed

types/core.d.ts
sarveshnagpal: Named it RegionTracker #Closed

src/layout/mutation.ts
sarveshnagpal: Called it extractRegions. #Closed

src/layout/dom.ts
sarveshnagpal: Yes, the value can be null but it is guaranteed to exist. #Closed

src/layout/dom.ts
sarveshnagpal: Named it data-clarity-region and CLARITY_REGION_ATTRIBUTE #Closed

src/data/metadata.ts
sarveshnagpal: Fixed, thanks. #Closed

src/clarity.ts
toby-walker: It's called consent() in latest version. Can you change it t a cleaere name than just "track" if consent is not clear? For example enableTracking() or startTracking() or something. Your concise names are sometimes a bit cryptic. :-)

---
In reply to: [396137009](https://github.com/microsoft/clarity-js/pull/170#discussion_r396137009) [](ancestors = 396137009)

src/layout/dom.ts
toby-walker: Please add above as comment. Not obvious that is why you need to check each case. 

---
In reply to: [395765133](https://github.com/microsoft/clarity-js/pull/170#discussion_r395765133) [](ancestors = 395765133)

src/layout/dom.ts
toby-walker: What does it mean to track a parent node and not its children?

---
In reply to: [395788883](https://github.com/microsoft/clarity-js/pull/170#discussion_r395788883) [](ancestors = 395788883)

toby-walker (changes requested):
:clock1:

src/layout/dom.ts
sarveshnagpal: Done #Closed

toby-walker (approved):
:shipit:

PR #172, opened 2020-04-06
Bug fixes and improvements in v1.0.0-beta.21

sarveshnagpal:
1) Support for shadow DOM
2) Adding support for backward & forward compat
3) Auto mask sensitive fields
4) Fix a bug where first payload was missing


decode/clarity.ts
```diff
@@ -26,8 +26,18 @@ export function decode(input: string): DecodedPayload {
...
+    // We do an exact match for major, minor and path components of the version.
+    // However, the beta portion of the version can be either same, one less or one more.
+    // This ensures we are backward and forward compatible with upto one version change.
+    let jsonVersion = parseVersion(payload.envelope.version);
+    let codeVersion = parseVersion(version);
+
+    if (jsonVersion.major !== codeVersion.major ||
+        jsonVersion.minor !== codeVersion.minor ||
+        jsonVersion.patch !== codeVersion.patch ||
+        Math.abs(jsonVersion.beta - codeVersion.beta) > 1) {
```
toby-walker, line 39: You should change this to be more compartible with semantic versioning. So if the major version is the same and the minor version is the same or higher you should decode it. 
It would be good to start avoiding the b21 type numbers and just use 0.21.0 or something similar (where 0 is major, 21 is minor and 0 is patch).  #Closed

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clarity-js",
-  "version": "1.0.0-beta.20",
+  "version": "1.0.0-beta.21",
```
toby-walker, line 3: Minor but I think we should change this and make it something more like 0.21.0 to be more consistent with semantic versioning. Using hte beta to discimrinate our version puts the emphasis on the least used part of semantic versioniong. 

From:https://semver.org/
Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.
Version 1.0.0 defines the public API. The way in which the version number is incremented after this release is dependent on this public API and how it changes. #Closed

decode/render.ts
```diff
@@ -113,12 +115,12 @@ function css(style: CSSStyleDeclaration, field: string): number {
 
 export function markup(type: Event, data: DomData[], iframe: HTMLIFrameElement): void {
     let doc = iframe.contentDocument;
-    if (type === Event.Discover) { reset(); }
     for (let node of data) {
         let parent = element(node.parent);
         let next = element(node.next);
         switch (node.tag) {
             case "*D":
```
toby-walker: What do these cases mean and what are the tags from? Could you make these more meaningfully named constants? #Closed

decode/render.ts
```diff
@@ -113,12 +115,12 @@ function css(style: CSSStyleDeclaration, field: string): number {
 
 export function markup(type: Event, data: DomData[], iframe: HTMLIFrameElement): void {
     let doc = iframe.contentDocument;
-    if (type === Event.Discover) { reset(); }
```
toby-walker, line 116: Is this the change for the first payload missing? #Closed

toby-walker:
>     let el = element(id) as HTMLElement;

Where is this defined? I couldn't see it anywhere in the code.

---
Refers to: decode/render.ts:84 in deb33bf. [](commit_id = deb33bf13edb51b9b4795e4aa03ca7e804634153, deletion_comment = False)

decode/render.ts
```diff
@@ -277,7 +295,10 @@ export function change(data: InputChangeData, iframe: HTMLIFrameElement): void {
 export function selection(data: SelectionData, iframe: HTMLIFrameElement): void {
     let doc = iframe.contentDocument;
     let s = doc.getSelection();
-    s.setBaseAndExtent(element(data.start as number), data.startOffset, element(data.end as number), data.endOffset);
+    // Wrapping selection code inside a try / catch to avoid throwing errors when dealing with elements inside the shadow DOM.
```
toby-walker, line 318: >selection [](start = 16, length = 9)

Nice comment. Could you expand what it means to have a selection in the shadow DOM? In particular can a selection include only a part of the shadow DOM or is always going to include  the entire shadow DOM (i.e. the element that the shadow DOM represents.

Could you also add this reference: 
https://developers.google.com/web/updates/2019/02/constructable-stylesheets

It's a bit more reader friendly (the spec is also great so having both here).  #Closed

src/core/mask.ts
```diff
@@ -1,14 +1,3 @@
 export default function(value: string): string {
-    let wasWhiteSpace = false;
-    let textCount = 0;
-    let wordCount = 0;
-    for (let i = 0; i < value.length; i++) {
-        let code = value.charCodeAt(i);
-        let isWhiteSpace = (code === 32 || code === 10 || code === 9 || code === 13);
-        let isNotCharacter = ((code >= 33 && code <= 47) || (code >= 91 && code <= 96) || (code >= 123 && code <= 126));
-        textCount += isWhiteSpace || isNotCharacter ? 0 : 1;
```
toby-walker, line 9: This looks like a pretty important change. Can you describe why you made it?
Also is this change backwards/forwards compatible for rendering? #Closed

src/data/upload.ts
```diff
@@ -122,10 +123,15 @@ export function end(): void {
...
 
-function upload(last: boolean = false): void {
+function upload(final: boolean = false): void {
     memory.compute();
     target.compute();
     metric.compute();
+
+    // Treat this as the last payload only if final boolean was explictly set to true
+    // In some edge cases, certain third party scripts (e.g. script error monitoring) could inject bogus arguments in the end to capture
+    // valid stack traces. In those cases, we don't want a random injected object (e.g. { "key": "value" }) to be mistaken for true.
```
toby-walker: >those [](start = 30, length = 5)

Can you expand a bit to describe what "inject bogus arguments" means and how it relates to capturing valid stack traces. Do you have an example of such a third party script?  #Closed

src/data/upload.ts
```diff
@@ -143,25 +149,27 @@ function stringify(payload: EncodedPayload): string {
...
 }
 
-function send(data: string, sequence: number = null, last: boolean = false): void {
+function send(data: string, sequence: number, last: boolean): void {
     // Upload data if a valid URL is defined in the config
     if (config.url.length > 0) {
-        if (last && "sendBeacon" in navigator) {
+        // Upload data using sendBeacon only when it's the last payload and the data size is within safe limits.
+        // Most browsers in the wild will reject the sendBeacon call if data being transferred is over 64KB (BEACON_BYTES_LIMIT).
+        if (last && data.length < BEACON_BYTES_LIMIT && "sendBeacon" in navigator) {
```
toby-walker: >BEACON_BYTES_LIMIT [](start = 34, length = 18)

Will the beacon call get an error or just be silently ignored?  #Closed

src/data/upload.ts
```diff
@@ -143,25 +149,27 @@ function stringify(payload: EncodedPayload): string {
...
             navigator.sendBeacon(config.url, data);
```
toby-walker: The second beacon call returns a Boolean if it succeeds. Could you change this to use that value and then fall back to the XHR if it return false? That seems safer than a hardcoded assumption of data size. See: return value in: [https://www.w3.org/TR/beacon/#sec-sendBeacon-method](https://www.w3.org/TR/beacon/#sec-sendBeacon-method). 
 #Closed

src/interaction/index.ts
```diff
@@ -1,3 +1,4 @@
+import * as change from "@src/interaction/change";
```
toby-walker: Will this enable input events? #Closed

types/decode/decode.d.ts
```diff
@@ -43,3 +43,10 @@ export interface DecodedPayload {
     network?: NetworkEvent[];
     paint?: PaintEvent[];
 }
+
+export interface DecodedVersion {
+    major: number;
+    minor: number;
+    patch: number;
+    beta: number;
```
toby-walker, line 53: Should change this to be prerelease. See general comment elsewhere about use of semver in this package. Happy to chat.  #Closed

src/layout/node.ts
```diff
@@ -17,29 +17,44 @@ export default function(node: Node, source: Source): void {
...
             let doctype = node as DocumentType;
             let docAttributes = { name: doctype.name, publicId: doctype.publicId, systemId: doctype.systemId };
             let docData = { tag: "*D", attributes: docAttributes };
-            dom[call](node, docData, source);
+            dom[call](node, parent, docData, source);
+            break;
+        case Node.DOCUMENT_FRAGMENT_NODE:
+            let shadowRoot = (node as ShadowRoot);
+            if (shadowRoot.host) {
+                // See: https://wicg.github.io/construct-stylesheets/
```
toby-walker: Very helpful comment (with ref). #Closed

src/layout/node.ts
```diff
@@ -17,29 +17,44 @@ export default function(node: Node, source: Source): void {
...
+                // For more details on adoptedStyleSheets for shadow dom
+                let style = "";
+                let adoptedStyleSheets: CSSStyleSheet[] = "adoptedStyleSheets" in shadowRoot ? shadowRoot["adoptedStyleSheets"] : [];
+                for (let styleSheet of adoptedStyleSheets) { style += getCssRules(styleSheet); }
+                let fragementData = { tag: "*S", attributes: { style } };
```
toby-walker: >"*S" [](start = 43, length = 4)

See comment elsewhere but just give these constant string names.  #Closed

src/layout/node.ts
```diff
@@ -17,29 +17,44 @@ export default function(node: Node, source: Source): void {
...
     switch (node.nodeType) {
         case Node.DOCUMENT_TYPE_NODE:
             let doctype = node as DocumentType;
             let docAttributes = { name: doctype.name, publicId: doctype.publicId, systemId: doctype.systemId };
             let docData = { tag: "*D", attributes: docAttributes };
-            dom[call](node, docData, source);
+            dom[call](node, parent, docData, source);
+            break;
+        case Node.DOCUMENT_FRAGMENT_NODE:
+            let shadowRoot = (node as ShadowRoot);
```
toby-walker, line 36: Is there any chance this might not be a Shadow DOM node and the shadowRoot would be null? Might be worth changing condition below to if(shadowRoot && )
Also consider updating to TypeScript 3.7 as it finally has option chanhing (.?) to avoid all the if(x && x.foo) or x == null ? x.foo : null junk.
See: [https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html) #Closed

src/layout/node.ts
```diff
@@ -17,29 +17,44 @@ export default function(node: Node, source: Source): void {
...
+            if (shadowRoot.host) {
+                // See: https://wicg.github.io/construct-stylesheets/
+                // For more details on adoptedStyleSheets for shadow dom
+                let style = "";
+                let adoptedStyleSheets: CSSStyleSheet[] = "adoptedStyleSheets" in shadowRoot ? shadowRoot["adoptedStyleSheets"] : [];
```
toby-walker: >adoptedStyleSheets [](start = 59, length = 18)

This is only supported in recent Chromium browser ([https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot](https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot)). What happens to other ones Safari/Firefox/etc.? #Closed

src/layout/node.ts
toby-walker: >tag [](start = 38, length = 3)

Are these tags sent over the wire or only used internally? #Closed

src/layout/mutation.ts
```diff
@@ -37,9 +36,15 @@ export function start(): void {
     };
 }
 
+export function observe(node: Node): void {
+  let observer = window["MutationObserver"] ? new MutationObserver(measure(handle) as MutationCallback) : null;
```
toby-walker: Why do you create a new observer each time here rather than once (as before the change)? Would be great to add a comment with your explanation.  #Closed

src/layout/index.ts
```diff
@@ -1,19 +1,17 @@
...
 import * as doc from "@src/layout/document";
 import * as dom from "@src/layout/dom";
 import * as mutation from "@src/layout/mutation";
 
 export function start(): void {
     doc.reset();
-    dom.reset();
+    dom.start();
     mutation.start();
-    discover.start();
```
toby-walker, line 11: What's the semantics of star/tend/reset and why aren't they all the same here in the start/end methods?  #Pending

src/layout/dom.ts
```diff
@@ -59,23 +78,29 @@ export function getId(node: Node, autogen: boolean = false): number {
...
-        masked = parent.metadata.masked;
+        parentValue = values[parentId];
+        parentValue.children.push(id);
+        region = region === null ? parentValue.region : region;
+        masked = parentValue.metadata.masked;
     }
 
-    if (data.attributes && Constant.MASK_ATTRIBUTE in data.attributes) { masked = true; }
-    if (data.attributes && Constant.UNMASK_ATTRIBUTE in data.attributes) { masked = false; }
+    // If element has a valid shadowRoot, track Shadow DOM as a top level root node.
```
toby-walker: What happens if the shadow DOM is attached in closed rather than open mode? In that case shadowRoot will be null. Will that affect our ability to playback the page? Should we check for that case here explicitly and send back an indication that there was a shadom that was attached in closed mode and, therefore, something could not be captured?

Should we monkey patch attachShadow to force everything to be open? See https://blog.revillweb.com/open-vs-closed-shadow-dom-9f3d7427d1af #Closed

src/layout/dom.ts
```diff
@@ -141,6 +167,9 @@ export function update(node: Node, data: NodeInfo, source: Source): void {
             }
         }
 
+        // If element has a valid shadowRoot, track Shadow DOM as a top level root node.
+        if ("shadowRoot" in element && element.shadowRoot && !has(element.shadowRoot)) { discoverTree(element.shadowRoot); }
```
toby-walker: Very minor; Might move this logic to a separate function to avoid repeating it a few places.  #Closed

src/layout/dom.ts
```diff
@@ -156,6 +185,34 @@ export function update(node: Node, data: NodeInfo, source: Source): void {
...
+    let attributes = data.attributes;
+    let tag = data.tag.toUpperCase();
+
+    // Do not proceed if attributes are missing for the node
+    if (attributes === null || attributes === undefined) { return masked; }
+
+    // Check for blacklist fields (e.g. address, phone, etc.) only if the input node is not already masked
+    if (masked === false && tag === Constant.TAG_INPUT && Constant.NAME_ATTRIBUTE in attributes) {
+        let value = attributes[Constant.NAME_ATTRIBUTE].toLowerCase();
+        for (let name of BLACKLIST_NAMES) {
```
toby-walker: Should we do this using CSS selectors instead? Would give us the option of configuring masking by CSS Selectors as a general feature for later like other system provide and should allow everything you do in a simpler and more general way. 
This is an interesting model: [https://help.fullstory.com/hc/en-us/articles/360044349073-FullStory-Private-by-Default](https://help.fullstory.com/hc/en-us/articles/360044349073-FullStory-Private-by-Default) #Closed

toby-walker: >BLACKLIST [](start = 25, length = 9)

Change the term "blacklist" to "disallowed list" or similar. Not considered a naming "best practice" anymore.  #Closed

src/layout/dom.ts
```diff
@@ -1,9 +1,14 @@
+import { Priority } from "@clarity-types/core";
 import { Constant, NodeChange, NodeInfo, NodeValue, Source } from "@clarity-types/layout";
 import config from "@src/core/config";
+import { schedule } from "@src/core/task";
 import time from "@src/core/time";
+import discover from "@src/layout/discover";
 import selector from "@src/layout/selector";
 
 let index: number = 1;
+const BLACKLIST_TYPES = ["password", "hidden", "email"];
```
toby-walker: Can you document the source of these lists with a reference?  #Closed

toby-walker: This list might need to large and add: tel, email

ee : [https://www.w3schools.com/html/html_form_input_types.asp](https://www.w3schools.com/
html/html_form_input_types.asp)
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input#%3Cinput%3E_types
 #Closed

src/layout/dom.ts
```diff
@@ -1,9 +1,14 @@
...
+const BLACKLIST_NAMES = ["address", "cell", "code", "dob", "email", "mobile", "name", "phone", "secret", "social", "ssn", "tel", "zip"];
```
toby-walker: This is just a heuristic, correct? None of these has any "official" meaning.  #Closed

decode/clarity.ts
sarveshnagpal: We will need to go with 1.X.Y right now. See here: https://www.npmjs.com/package/clarity-js?activeTab=versions - we already used 0.X.Y for previous clarity-js. The idea was that we will go to normal usage once we get out of beta. May be now is the time and I should snap sarveshn/vnext back to master and stop publishing beta versions? We will sync more offline. #Closed

decode/render.ts
sarveshnagpal: This is playback code. Earlier discover event was sent only once and whenever we received it, it would mean we are dealing a new impression (mostly for extension). With introduction of shadow dom it no longer is true so I removed that condition.  #Closed

decode/render.ts
sarveshnagpal: No, this is all playback (not mainline instrumentation code). See comment above for explanation. #Closed

src/data/upload.ts
sarveshnagpal: We invoke beacon call when we are almost ready to leave the page - on unload. At that point we can't wait and then reissue the request because browser would've already unloaded us from the memory by then. We have one chance to fire the request and if we don't get it right, we lost the chance. I could go with sync xhr but that can have performance implications so not going there and keeping it simple. #Closed

src/interaction/index.ts
sarveshnagpal: Yes, it was a miss from many version before. It wires it up now. #Closed

types/decode/decode.d.ts
sarveshnagpal: Yes will do prerelease and will chat offline. Thanks. #Closed

src/layout/node.ts
sarveshnagpal: Sent over the wire #Closed

src/layout/mutation.ts
sarveshnagpal: I'll add the comment. Basically each dom tree within shadow DOM needs a different observer since uber observer no longer works within shadow DOM.  #Closed

src/data/upload.ts
sarveshnagpal: A sample example is: https://www.npmjs.com/package/raven-js #Closed

src/data/upload.ts
sarveshnagpal: Silently ignored. In the network panel, browser doesn't even put this on the network and it keeps showing as (pending).  #Closed

src/layout/node.ts
sarveshnagpal: They are sent over the wire and follow the same pattern as previous tags e.g. *D, *B. I'll make a const enum so it's easier to follow them while saving bytes over the wire.  #Closed

src/layout/node.ts
sarveshnagpal: shadowRoot is just a typed version of node for Typescript. So if node is null, we wouldn't be able to get inside the switch / case anyway because the uber switch condition is node.nodeType. #Closed

src/layout/node.ts
sarveshnagpal: Shadow DOM is supported in many browsers (https://caniuse.com/#search=shadow) however adoptedStyleSheets (which is a way for web components to efficiently share CSS) is only supported in select browsers. Where it is not supported, they will be using classic style tag for styles and we already support that. Also for playback, we don't rely on adoptedStyleSheets since playback could be another browser and we turn these into regular style sheets (see render.ts). In short, we should be good cross browser. #Closed

src/layout/mutation.ts
sarveshnagpal: Added a comment #Closed

src/layout/dom.ts
sarveshnagpal: If shadow dom is closed then yes we can't access (similar to third party IFRAME). Should we monkey patch? Not in favor - at least not yet. I think it's ok for Clarity to proxy the calls but not alter the calls. For instance if a web component really wanted a closed shadow dom - we should continue with that and not break the page (unintentionally by changing it to open). Also, once this problem happens in practice, we can revisit it. I have not seen any real use case for open shadow dom or encountered one yet.  #Closed

decode/clarity.ts
sarveshnagpal: Synced offline. Closing this for now. We will go with 1.X.Y when time is right. #Closed

decode/render.ts
sarveshnagpal: Added the comment in main instrumentation code per our offline discussion.  #Closed

src/data/upload.ts
sarveshnagpal: Fixed as you suggested based on our offline conversation. #Closed

types/decode/decode.d.ts
sarveshnagpal: Not changing this but eventually move to more reasonable versioning soon. #Closed

src/layout/node.ts
sarveshnagpal: Changed into const enums #Closed

src/layout/index.ts
sarveshnagpal: I'll come back to it later. In some cases there's no need to do anything in the end function so i have left it out, instead of adding bogus bytes. I will reconsider this later and see if it makes sense to reintroduce for consistency anyway. #Pending

src/layout/dom.ts
sarveshnagpal: I simplified this so it won't look as bad.  #Closed

src/layout/dom.ts
sarveshnagpal: Will do so in subsequent release.  #Closed

toby-walker: Simple name change. Can you just does this now before you check in?

---
In reply to: [405123883](https://github.com/microsoft/clarity-js/pull/172#discussion_r405123883) [](ancestors = 405123883)

src/interaction/encode.ts
```diff
@@ -52,14 +62,15 @@ export default async function(type: Event): Promise<void> {
...
             break;
-        case Event.InputChange:
-            let ch = change.data;
-            if (ch) {
-                tokens.push(observe(ch.target as TargetInfo));
-                tokens.push(ch.value);
-                change.reset();
+        case Event.Input:
+            for (let i = 0; i < input.state.length; i++) {
+                let entry = input.state[i];
```
toby-walker, line 67: Why are you iterating over a list of things for the state here? #Closed

src/core/mask.ts
sarveshnagpal: This was premature optimization. For now I've kept things simple and we can worry about optimizations later. This only impacts lean mode. #Closed

src/layout/dom.ts
sarveshnagpal: Synced offline. #Closed

sarveshnagpal: I've added tel. #Closed

src/layout/dom.ts
sarveshnagpal: Yes right.

src/interaction/encode.ts
sarveshnagpal: We look ahead for 500ms to receive input. If we get input events for two fields (someone typing fast, etc.) then we will queue up two different input events (one for each individual target). This loop helps us ensure we send all events.  #Closed

src/layout/dom.ts
sarveshnagpal: Done, missed it before. #Closed

toby-walker (approved):
:shipit:

src/layout/dom.ts
toby-walker: Please add comment about that.

---
In reply to: [406557724](https://github.com/microsoft/clarity-js/pull/172#discussion_r406557724) [](ancestors = 406557724)

PR #173, opened 2020-04-20
Bug fixes and improvements in v1.0.0-beta.22

sarveshnagpal:
- Adding support for same-origin iframes

### DIFF
--- a/decode/render.ts
+++ b/decode/render.ts
@@ -52,12 +52,12 @@ export function metric(data: MetricData, header: HTMLElement): void {
         if (metrics[entry]) {
             let m = metrics[entry];
             let map = METRIC_MAP[entry];
-            let unit = "unit" in map ? map.unit : "";
+            let unit = "unit" in map ? map.unit : Constant.EMPTY_STRING;
             html.push(`<li><h2>${value(m, unit)}<span>${unit}</span></h2>${map.name}</li>`);
         }
     }
 
-    header.innerHTML = `<ul>${html.join("")}</ul>`;
+    header.innerHTML = `<ul>${html.join(Constant.EMPTY_STRING)}</ul>`;
 }
 
 function value(num: number, unit: string): number {
@@ -106,33 +106,18 @@ export function markup(type: Event, data: DomData[], iframe: HTMLIFrameElement):
         let tag = node.tag && node.tag.indexOf(Constant.IFRAME_PREFIX) === 0 ? node.tag.substr(Constant.IFRAME_PREFIX.length) : node.tag;
         switch (tag) {
             case Constant.DOCUMENT_TAG:
-                if (type === Event.Discover) { reset(); }
-                if (typeof XMLSerializer !== "undefined") {
-                    doc.open();
-                    doc.write(new XMLSerializer().serializeToString(
-                        doc.implementation.createDocumentType(
+                let tagDoc = tag !== node.tag ? (parent ? (parent as HTMLIFrameElement).contentDocument : null): doc;
+                if (tagDoc && tagDoc === doc && type === Event.Discover) { reset(); }
+                if (typeof XMLSerializer !== "undefined" && tagDoc) {
+                    tagDoc.open();
+                    tagDoc.write(new XMLSerializer().serializeToString(
+                        tagDoc.implementation.createDocumentType(
                             node.attributes["name"],
                             node.attributes["publicId"],
                             node.attributes["systemId"]
                         )
                     ));
-                    doc.close();
-                }
-                break;
-            case Constant.FRAME_DOCUMENT_TAG:
-                if (typeof XMLSerializer !== "undefined" && parent) {
-                    let frame = parent as HTMLIFrameElement;
-                    if (frame.contentDocument) {
-                        frame.contentDocument.open();
-                        frame.contentDocument.write(new XMLSerializer().serializeToString(
-                            frame.contentDocument.implementation.createDocumentType(
-                                node.attributes["name"],
-                                node.attributes["publicId"],
-                                node.attributes["systemId"]
-                            )
-                        ));
-                        frame.contentDocument.close();
-                    }
+                    tagDoc.close();
                 }
                 break;
             case Constant.POLYFILL_SHADOWDOM_TAG:
@@ -166,19 +151,19 @@ export function markup(type: Event, data: DomData[], iframe: HTMLIFrameElement):
                 insert(node, parent, textElement, next);
                 break;
             case "HTML":
-                let d = tag !== node.tag ? (parent ? (parent as HTMLIFrameElement).contentDocument : null): doc;
-                if (d !== null) {
+                let htmlDoc = tag !== node.tag ? (parent ? (parent as HTMLIFrameElement).contentDocument : null): doc;
+                if (htmlDoc !== null) {
                     let docElement = element(node.id);
                     if (docElement === null) {
-                        let newDoc = d.implementation.createHTMLDocument("");
+                        let newDoc = htmlDoc.implementation.createHTMLDocument(Constant.EMPTY_STRING);
                         docElement = newDoc.documentElement;
-                        let p = d.importNode(docElement, true);
-                        d.replaceChild(p, d.documentElement);
-                        if (d.head) { d.head.parentNode.removeChild(d.head); }
-                        if (d.body) { d.body.parentNode.removeChild(d.body); }
+                        let p = htmlDoc.importNode(docElement, true);
+                        htmlDoc.replaceChild(p, htmlDoc.documentElement);
+                        if (htmlDoc.head) { htmlDoc.head.parentNode.removeChild(htmlDoc.head); }
+                        if (htmlDoc.body) { htmlDoc.body.parentNode.removeChild(htmlDoc.body); }
                     }
-                    setAttributes(d.documentElement as HTMLElement, node.attributes);
-                    nodes[node.id] = d.documentElement;
+                    setAttributes(htmlDoc.documentElement as HTMLElement, node.attributes);
+                    nodes[node.id] = htmlDoc.documentElement;
                 }
                 break;
             case "HEAD":
--- a/decode/render.ts
+++ b/decode/render.ts
@@ -103,7 +103,8 @@ export function markup(type: Event, data: DomData[], iframe: HTMLIFrameElement):
     for (let node of data) {
         let parent = element(node.parent);
         let next = element(node.next);
-        switch (node.tag) {
+        let tag = node.tag && node.tag.indexOf(Constant.IFRAME_PREFIX) === 0 ? node.tag.substr(Constant.IFRAME_PREFIX.length) : node.tag;
+        switch (tag) {
             case Constant.DOCUMENT_TAG:
                 if (type === Event.Discover) { reset(); }
                 if (typeof XMLSerializer !== "undefined") {
@@ -165,13 +166,7 @@ export function markup(type: Event, data: DomData[], iframe: HTMLIFrameElement):
                 insert(node, parent, textElement, next);
                 break;
             case "HTML":
-                let htmlElement = element(node.id) as HTMLElement;
-                // If it's a mutation to remove the HTML node from the page, handle it like any other DOM element
-                if (htmlElement && parent === null) {
-                    insert(node, parent, htmlElement, next);
-                    break;
-                }
-                let d = parent ? (parent as HTMLIFrameElement).contentDocument : doc;
+                let d = tag !== node.tag ? (parent ? (parent as HTMLIFrameElement).contentDocument : null): doc;
                 if (d !== null) {
                     let docElement = element(node.id);
                     if (docElement === null) {
--- a/decode/render.ts
+++ b/decode/render.ts
@@ -94,23 +94,8 @@ function box(id: number, region: string, rectangle: number[], iframe: HTMLIFrame
             doc.body.appendChild(layer);
             layer.innerText = region;
             nodes[id] = layer;
-        } else if (el && el.tagName === "IFRAME") {
-            let s = getComputedStyle(el, null);
-            let width = rectangle[2];
-            let height = rectangle[3];
-            if (s["boxSizing"] !== "border-box") {
-                width -= (css(s, "paddingLeft") + css(s, "paddingRight") + css(s, "borderLeftWidth") + css(s, "borderRightWidth"));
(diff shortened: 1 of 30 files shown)


decode/render.ts
```diff
@@ -94,25 +94,10 @@ function box(id: number, region: string, rectangle: number[], iframe: HTMLIFrame
             doc.body.appendChild(layer);
             layer.innerText = region;
             nodes[id] = layer;
-        } else if (el && el.tagName === "IFRAME") {
-            let s = getComputedStyle(el, null);
-            let width = rectangle[2];
```
toby-walker, line 99: This is the change that removes iframes from the box model data and only sends for regions? #Closed

src/core/event.ts
```diff
@@ -12,7 +12,7 @@ export function bind(target: EventTarget, event: string, listener: EventListener
 export function reset(): void {
   // Walk through existing list of bindings and remove them all
   for (let binding of bindings) {
-    (binding.target).removeEventListener(binding.event, binding.listener, binding.capture);
+    try { binding.target.removeEventListener(binding.event, binding.listener, binding.capture); } catch { /* do nothing */ }
```
toby-walker: Add a comment about why this could fail. Also is this the only location you remove event listeners? If not, should other locations also get this error handling? #Closed

src/interaction/click.ts
```diff
@@ -3,18 +3,23 @@ import { ClickData } from "@clarity-types/interaction";
...
-function handler(event: Event, evt: MouseEvent): void {
-    let de = document.documentElement;
-    let x = "pageX" in evt ? Math.round(evt.pageX) : ("clientX" in evt ? Math.round(evt["clientX"] + de.scrollLeft) : null);
-    let y = "pageY" in evt ? Math.round(evt.pageY) : ("clientY" in evt ? Math.round(evt["clientY"] + de.scrollTop) : null);
+function handler(event: Event, root: Node, evt: MouseEvent): void {
+    let frame = iframe(root);
+    let d = frame ? frame.contentDocument.documentElement : document.documentElement;
+    let x = "pageX" in evt ? Math.round(evt.pageX) : ("clientX" in evt ? Math.round(evt["clientX"] + d.scrollLeft) : null);
+    let y = "pageY" in evt ? Math.round(evt.pageY) : ("clientY" in evt ? Math.round(evt["clientY"] + d.scrollTop) : null);
+    x = x && frame ? x + frame.offsetLeft : x;
```
toby-walker, line 25: Minor; Add comment about why need to offset coordinates by iframe top left here. #Closed

src/interaction/index.ts
```diff
@@ -8,9 +8,9 @@ import * as unload from "@src/interaction/unload";
 import * as visibility from "@src/interaction/visibility";
 
 export function start(): void {
-    click.start();
-    input.start();
```
toby-walker, line 12: The different between reset and start here is very a hard to understand.

src/interaction/index.ts
```diff
@@ -19,8 +19,22 @@ export function start(): void {
 }
 
 export function end(): void {
+    click.reset();
+    pointer.end();
     input.end();
     pointer.end();
     scroll.end();
     selection.end();
```
toby-walker, line 28: Why are some thing not present here like unload.end()?

src/interaction/index.ts
```diff
@@ -19,8 +19,22 @@ export function start(): void {
...
 }
+
+export function observe(root: Node): void {
+    scroll.observe(root);
+    // Only monitor following interactions if the root node is a document
+    // In case of shadow DOM, following events automatically bubble up to the parent document.
```
toby-walker, line 35: Will these event inside the Shadow DOM be exposed here?

src/interaction/scroll.ts
```diff
@@ -19,7 +20,8 @@ export function start(): void {
 }
 
 export function observe(root: Node): void {
-    let node = root === document ? window : root;
+    let frame = iframe(root);
+    let node = frame ? frame.contentWindow : (root === document ? window : root);
```
toby-walker, line 24: >contentWindow [](start = 29, length = 13)

Why contentWindow and not contentDocument?

src/interaction/selection.ts
```diff
@@ -13,12 +13,16 @@ let timeout: number = null;
...
 
-function recompute(): void {
-    let current = document.getSelection();
+export function observe(root: Node): void {
+    bind(root, "selectstart", recompute.bind(this, root), true);
+    bind(root, "selectionchange", recompute.bind(this, root), true);
+}
+
+function recompute(root: Node): void {
+    let doc = root.nodeType === Node.DOCUMENT_NODE ? root as Document : document;
```
toby-walker, line 24: This is window.document? Would be clearer if written that way (to me0. #Closed

src/layout/dom.ts
```diff
@@ -328,9 +350,6 @@ function metadata(tag: string, id: number, parentId: number): void {
                     }
                 }
                 break;
-            case "IFRAME":
-                if (config.lean === false) { value.metadata.boxmodel = true; }
```
toby-walker, line 332: Is this the fix to not send iframes in box model? #Closed

src/layout/dom.ts
```diff
@@ -171,6 +173,26 @@ export function update(node: Node, parent: Node, data: NodeInfo, source: Source)
     }
 }
 
+export function sameorigin(node: Node): boolean {
+    let output = false;
+    if (node.nodeType === Node.ELEMENT_NODE && (node as HTMLElement).tagName === Constant.IFRAME_TAG) {
+        let frame = node as HTMLIFrameElement;
```
toby-walker, line 179: >HTMLIFrameElement [](start = 28, length = 17)

When could this error happen? Add a brief comment. I assume it is in cases where not same origin causing a security exception? #Closed

src/layout/dom.ts
```diff
@@ -171,6 +173,26 @@ export function update(node: Node, parent: Node, data: NodeInfo, source: Source)
...
+                output = true;
+            }
+        } catch { /* do nothing */ }
+    }
+    return output;
+}
+
+export function iframe(node: Node): HTMLIFrameElement {
+    let doc = node.nodeType === Node.DOCUMENT_NODE ? node as Document : null;
+    return doc && iframeMap.has(doc) ? iframeMap.get(doc) : null;
```
toby-walker, line 196: >iframeMap [](start = 18, length = 9)

Minor suggestion: Maybe iframeMap -. iframeDoc2FrameMap to indicate what the mapping is from/to in the name? #Closed

src/layout/mutation.ts
```diff
@@ -50,6 +51,15 @@ export function observe(node: Node): void {
   } catch (error) { internal.error(Code.MutationObserver, error); }
 }
 
+export function monitor(frame: HTMLIFrameElement): void {
+  // Bind to iframe's onload event so we get notified anytime there's an update to iframe contents.
```
toby-walker: >contents [](start = 90, length = 8)

contents -> content #Closed

src/layout/node.ts
```diff
@@ -20,17 +22,23 @@ export default function(node: Node, source: Source): void {
...
+    let add = dom.has(node) === false;
+    let call = add ? "add" : "update";
     let parent = node.parentElement ? node.parentElement : null;
+    let insideFrame = node.ownerDocument !== document;
     switch (node.nodeType) {
-        case Node.DOCUMENT_NODE:
-            observe(node);
-            break;
         case Node.DOCUMENT_TYPE_NODE:
+            // Skip processing nodes within iframe if they were already removed from the DOM before we got to processing them.
```
toby-walker: How could this happen?

src/layout/node.ts
```diff
@@ -20,17 +22,23 @@ export default function(node: Node, source: Source): void {
...
+            // In those cases, we skip them and continue processing rest of the nodes.
+            let frame = insideFrame && node.parentNode ? dom.iframe(node.parentNode) : null;
```
toby-walker: What happens with multiple nested iframes? That is, if an iframe contains an iframe contanis an iframe. Do we want ot handle that and do the changes you added handle it?

src/layout/node.ts
```diff
@@ -20,17 +22,23 @@ export default function(node: Node, source: Source): void {
...
+            if (insideFrame && frame === null) { break; }
             let doctype = node as DocumentType;
             let docAttributes = { name: doctype.name, publicId: doctype.publicId, systemId: doctype.systemId };
-            let docData = { tag: Constant.DOCUMENT_TAG, attributes: docAttributes };
-            dom[call](node, parent, docData, source);
+            let docData = { tag: insideFrame ? Constant.FRAME_DOCUMENT_TAG : Constant.DOCUMENT_TAG, attributes: docAttributes };
```
toby-walker: What does this value represent? #Closed

src/layout/node.ts
```diff
@@ -74,36 +82,58 @@ export default function(node: Node, source: Source): void {
...
+                    break;
                 case "SCRIPT":
                 case "NOSCRIPT":
                 case "META":
                     break;
                 case "HEAD":
                     let head = { tag, attributes: getAttributes(element.attributes) };
                     // Capture base href as part of discovering DOM
-                    head.attributes[Constant.BASE_TAG] = location.protocol + "//" + location.hostname;
+                    if (insideFrame === false) { head.attributes[Constant.BASE_ATTRIBUTE] = location.protocol + "//" + location.hostname; }
```
toby-walker, line 97: >) [](start = 45, length = 1)

Why not capture this in an iframe if it exists?

src/layout/node.ts
```diff
@@ -74,36 +82,58 @@ export default function(node: Node, source: Source): void {
...
                     let styleData = { tag, attributes, value: getStyleValue(element as HTMLStyleElement) };
                     dom[call](node, parent, styleData, source);
                     break;
+                case "IFRAME":
+                    let iframe = node as HTMLIFrameElement;
+                    let frameData = { tag, attributes: getAttributes(iframe.attributes) };
+                    if (dom.sameorigin(iframe)) {
+                        mutation.monitor(iframe);
+                        frameData.attributes[Constant.SAME_ORIGIN_ATTRIBUTE] = "true";
+                        if (iframe.contentDocument && iframe.contentWindow && iframe.contentDocument.readyState !== "loading") {
```
toby-walker, line 111: >readyState [](start = 101, length = 10)

Is the idea if the state is loading then you defer scanning content of the iframe until you get the load event that you have listened to? Wouldn't that miss mutations as the iframe is in the process of loading?

toby-walker (changes requested):
:clock1:

decode/render.ts
sarveshnagpal: Yes that's right.  #Closed

src/core/event.ts
sarveshnagpal: Thanks, added a comment.  #Closed

sarveshnagpal: Also, added it to binding events and also added a comment. All event bindings go through this central part, so we should be ok.  #Closed

src/interaction/index.ts
sarveshnagpal: As captured in the comment, shadow dom already bubbles up some of the events to parent DOM so we don't have to rewire them up.

src/interaction/scroll.ts
sarveshnagpal: We have been tracking scroll events on the window from before (see red part). Explanation is in the same file below in comments:
  // Edge doesn't support scrollTop position on document.documentElement.
  // For cross browser compatibility, looking up pageYOffset on window if the scroll is on document.

src/interaction/selection.ts
sarveshnagpal: This could also be iframe.contentDocument too and that's why we are checking if we should bind to a different document, and not the window.document. In other cases (e.g. shadowDom) we bind to "document" which is same as "window.document". #Closed

src/layout/dom.ts
sarveshnagpal: Yes, right.  #Closed

src/layout/dom.ts
sarveshnagpal: Yes thanks, will add comment. When it's cross-origin, browser will throw exception.  #Closed

src/layout/dom.ts
sarveshnagpal: Added in the comments to clarify. Not changing names to stay consistent with rest of the naming. #Closed

src/layout/mutation.ts
sarveshnagpal: Thanks, fixed. #Closed

src/interaction/click.ts
sarveshnagpal: Thanks done. #Closed

src/layout/node.ts
sarveshnagpal: I've simplified this to not use it anymore. I'm using prefix like we already use with SVG. Thanks for pointing out. #Closed

src/layout/node.ts
sarveshnagpal: We will capture it if it exists but we won't force generate the base tag. By default iframe will accquire parent's base tag so that works. However if IFRAME had it's own base, we would grab it like any other DOM element.

src/layout/node.ts
sarveshnagpal: Yes, we will miss the mutations but when we get to processing nodes within the IFRAME, we will get the current state at the point. This is true for the uber DOM too. If Clarity wires up late or we have to wait (due to async nature) and mutations happen inbetween, we merge them into initial discovery.

src/interaction/index.ts
sarveshnagpal: Agreed - fixed it. Now everything has a start & end. Leads to more bytes but we can optimize that later. Consistency is better for now.

src/interaction/index.ts
sarveshnagpal: Fixed it - made things consistent.

src/layout/node.ts
sarveshnagpal: Simplified this. Now handle it during rendering. For reference the way we would get into this state is: our code is async so by the time we are able to get to iframe, it's possible it may already have been destroyed.

src/layout/node.ts
sarveshnagpal: It works and yes the changes we have handle it. I've tried on few pages where we have iframe inside iframe inside iframe and so far haven't found issues in my testing.

